PR #34, opened 2021-02-25
Fix zero initializations

amklinv-nnl:


include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -398,7 +398,7 @@ void matrix_product(in_matrix_1_t A,
     const element_type alpha_A = extractScalingFactor(A, 1.0);
     const element_type alpha_B = extractScalingFactor(B, 1.0);
     const element_type alpha = alpha_A * alpha_B;
-    const element_type beta (0.0);
+    const element_type beta ();
```
mhoemmen, line 401: As discussed, please use `{}` instead of `()` here; thanks!

PR #35, opened 2021-02-26
Added continuous integration to the repo

amklinv-nnl:
Not perfect, but kinda good. Could definitely be better optimized.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,0 +1,79 @@
+name: CMake
+
+on: [push]
+
+env:
+  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
+  BUILD_TYPE: RelWithDebInfo
+
+jobs:
+  build:
+    # The CMake configure and build commands are platform agnostic and should work equally
+    # well on Windows or Mac.  You can convert this to a matrix build if you need
+    # cross-platform coverage.
+    # See: https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix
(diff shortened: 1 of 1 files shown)


.github/workflows/cmake.yml
```diff
@@ -0,0 +1,79 @@
...
+        repository: kokkos/mdarray
+        path: ${{github.workspace}}/mdarray
+      
+    - name: Configure CMake (mdarray)
+      shell: bash
+      working-directory: ${{github.workspace}}/build-mdarray
+      run: cmake $GITHUB_WORKSPACE/mdarray -DMDSpan_DIR=$GITHUB_WORKSPACE/build-mdspan -DCMAKE_BUILD_TYPE=$BUILD_TYPE
+
+    - name: Install gtest manually
+      run: sudo apt-get install libgtest-dev && cd /usr/src/gtest && sudo cmake CMakeLists.txt && sudo make && sudo cp *.a /usr/lib && sudo ln -s /usr/lib/libgtest.a /usr/local/lib/libgtest.a && sudo ln -s /usr/lib/libgtest_main.a /usr/local/lib/libgtest_main.a
```
mhoemmen, line 51: 1. Kokkos likes to build gtest too, at least it used to.  Is that OK with this CI testing?
2. Is the idea that these steps are supposed to execute automatically?  If so, then how can these `sudo` commands work?
3. Is there a way not to require `sudo`?

mhoemmen: Questions answered offline.  We can defer not installing googletest like this for later.

mhoemmen:
Thanks so much!!!  I very much look forward to CI testing! : - D

PR #36, opened 2021-03-03
Add gemv test

amklinv-nnl:


tests/gemv.cpp
```diff
@@ -0,0 +1,164 @@
...
+  using std::cout;
+  using std::endl;
+
+  template<class MdspanType, class Scalar>
+  struct FillMatrix {
+    static void fill(MdspanType A, const Scalar startVal);
+  };
+
+  template<class MdspanType>
+  struct FillMatrix<MdspanType, int> {
```
mhoemmen, line 24: This is perfectly fine, but I realize that a more modern way of doing this might be just by overloading, instead of this struct thing:
```cpp
template<class Layout, class Extents, class Accessor>
void fill(basic_mdspan<double, Layout, Extents, Accessor> A, const int startVal)
{ /* ... */ }
```

tests/gemv.cpp
```diff
@@ -0,0 +1,164 @@
...
+    using vector_t = basic_mdspan<scalar_t, extents<dynamic_extent>, layout_left>;
+
+    constexpr ptrdiff_t maxDim = 7;
+    constexpr ptrdiff_t storageSize(maxDim*maxDim + 3*maxDim);
+    std::vector<scalar_t> storage(storageSize);
+
+    for (ptrdiff_t numRows : {1, 4, 7}) {
+      for (ptrdiff_t numCols : {1, 4, 7}) {
+        ptrdiff_t offset = 0;
+        matrix_t A(storage.data() + offset, numRows, numCols);
```
mhoemmen, line 115: It's OK (but not necessary) to use mdarray here, if it's more convenient.

tests/gemv.cpp
```diff
@@ -0,0 +1,164 @@
...
+        vector_t y(storage.data() + offset, numRows);
+        offset += numRows;
+        vector_t gs(storage.data() + offset, numRows);
+
+        fill_matrix(A, scalar_t(real_t(1)));
+        fill_vector(x, scalar_t(real_t(2)));
+
+        // Initialize vector to zero
+        for (ptrdiff_t i = 0; i < numRows; i++) {
+          gs(i) = scalar_t(0.0); // this works even for complex
```
mhoemmen, line 128: It's fine for this test, but in general, `scalar_t{}` would do what you want without requiring that `0.0` be convertible to `scalar_t`.

tests/gemv.cpp
```diff
@@ -0,0 +1,164 @@
...
+        for (ptrdiff_t j = 0; j < numCols; ++j) {
+          for (ptrdiff_t i = 0; i < numRows; ++i) {
+            gs(i) += A(i,j) * x(j);
+          }
+        }
+
+        // Fill result vector with flag values to make sure that we
+        // computed everything.
+        for (ptrdiff_t j = 0; j < numRows; ++j) {
+          y(j) = std::numeric_limits<scalar_t>::min();
```
mhoemmen, line 141: For floating-point types, `min()` actually returns the smallest normalized positive value:

https://en.cppreference.com/w/cpp/types/numeric_limits/min

That's probably fine for this test -- it's just likely not what you meant.

mhoemmen (approved):
Thanks for adding the gemv test and adding PR testing as well!

PR #51, opened 2021-05-26
Create a sum of squares function

### LINKED ISSUE #43: Add vector_sum_of_squares
Add `vector_sum_of_squares`.  First, this gives implementers a path to implementing `vector_norm2` in a way that achieves the over/underflow guarantees intended by the BLAS Standard.  Second, this is a useful algorithm in itself for parallelizing vector 2-norm computation.

amklinv-nnl:
Replaces most of the vector_norm2 function. Potentially useful for other functions as well. Closes #43.

### DIFF
--- a/include/experimental/__p1673_bits/blas1_vector_sum_of_squares.hpp
+++ b/include/experimental/__p1673_bits/blas1_vector_sum_of_squares.hpp
@@ -40,10 +40,9 @@
 //@HEADER
 */
 
-#ifndef LINALG_INCLUDE_EXPERIMENTAL___P1673_BITS_BLAS1_VECTOR_NORM2_HPP_
-#define LINALG_INCLUDE_EXPERIMENTAL___P1673_BITS_BLAS1_VECTOR_NORM2_HPP_
+#ifndef LINALG_INCLUDE_EXPERIMENTAL___P1673_BITS_BLAS1_VECTOR_SUM_OF_SQUARES_HPP_
+#define LINALG_INCLUDE_EXPERIMENTAL___P1673_BITS_BLAS1_VECTOR_SUM_OF_SQUARES_HPP_
 
-#include "blas1_vector_sum_of_squares.hpp"
 #include <cmath>
 #include <cstdlib>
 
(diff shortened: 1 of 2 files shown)


include/experimental/__p1673_bits/blas1_vector_norm2.hpp
```diff
@@ -56,37 +57,19 @@ template<class in_vector_t,
 void vector_norm2(in_vector_t x,
                   Scalar& result)
```
mhoemmen, line 58: The proposal has us taking `T init` and returning `T`, just as `std::reduce` does.  We no longer take the result as as output argument.  I'll file a separate issue for that.

include/experimental/__p1673_bits/blas1_vector_norm2.hpp
```diff
@@ -56,37 +57,19 @@ template<class in_vector_t,
...
 {
-  using std::abs;
   using std::sqrt;
 
-  // Rescaling, as in the Reference BLAS DNRM2 implementation, avoids
-  // unwarranted overflow or underflow.
+  // Initialize the sum of squares result
+  sum_of_squares_result<Scalar> ssq_init;
```
mhoemmen, line 63: If `Scalar` is a floating-point type (`float`, `double`, or `long double`), you can do this in one line:
```
auto result = ssq_res = vector_sum_of_squares(x, {Scalar{}, Scalar{1}});
```
However, this brings up the interesting problem of what to do if `Scalar` is _not_ a floating-point type.  For instance, `Scalar` may be some type that doesn't know how to be initialized from `int` or `double`.  I think we can do this via `if constexpr`:
```
if constexpr(std::is_floating_point_v<Scalar>) {
  const auto ssq_res = vector_sum_of_squares(x, {Scalar{}, Scalar{1}});
  return ssq_res.scaling_factor * sqrt(ssq_res.scaled_sum_of_squares);
} else {
  Scalar result{};
  for(size_t k = 0; k < x.extent(0); ++k) {
    result += abs(x(i)) * abs(x(i));
  }
  return sqrt(result);
}
```

include/experimental/__p1673_bits/blas1_vector_norm2.hpp
mhoemmen: The separate issue: https://github.com/kokkos/stdBLAS/issues/52

include/experimental/__p1673_bits/blas1_vector_sum_of_squares.hpp
```diff
@@ -0,0 +1,114 @@
...
+  // unwarranted overflow or underflow.
+
+  Scalar scale = init.scaling_factor;
+  Scalar ssq = init.scaled_sum_of_squares;
+  for (ptrdiff_t i = 0; i < x.extent(0); ++i) {
+    if (abs(x(i)) != 0.0) {
+      const auto absxi = abs(x(i));
+      const auto quotient = scale / absxi;
+      if (scale < absxi) {
+          ssq = Scalar(1.0) + ssq * quotient * quotient;
```
mhoemmen, line 83: Again, we might need a separate implementation of `vector_sum_of_squares` for non-floating-point types, since we might not always be able to convert `1.0` to `Scalar`.

mhoemmen (approved):
Thanks @amklinv-nnl ! : - D  Please see comments for other issues.

PR #68, opened 2021-06-03
Enforce basic_mdspan-ness

### LINKED ISSUE #67: Make the implementation enforce basic_mdspan-ness
P1673's template parameter names like `in_vector_t` come with constraints (e.g., must be a rank-1 `basic_mdspan`).  The implementation needs to enforce that, by only overloading for `basic_mdspan` types instead of just templating on e.g., `in_vector_t`.  
For examples, see `vector_norm2` and `matrix_inf_norm`.

amklinv-nnl:
Closes #67

### DIFF
--- a/include/experimental/__p1673_bits/blas1_givens.hpp
+++ b/include/experimental/__p1673_bits/blas1_givens.hpp
@@ -345,14 +345,12 @@ label20:
   }
 }
 
-// FIXME (Hoemmen 2021/05/28) Latest version of P0009 (mdspan) uses size_t
-// instead of ptrdiff_t, but the implementation hasn't changed yet.
 template<class ElementType1,
-         ptrdiff_t ext1,
+         extents<>::size_type ext1,
          class Layout1,
          class Accessor1,
          class ElementType2,
-         ptrdiff_t ext2,
+         extents<>::size_type ext2,
          class Layout2,
          class Accessor2,
          std::floating_point Real>
@@ -362,7 +360,7 @@ void givens_rotation_apply(
   const Real c,
   const Real s)
 {
-  for (ptrdiff_t i = 0; i < x.extent(0); ++i) {
+  for (extents<>::size_type i = 0; i < x.extent(0); ++i) {
     const auto dtemp = c * x(i) + s * y(i);
     y(i) = c * y(i) - s * x(i);
     x(i) = dtemp;
@@ -371,11 +369,11 @@ void givens_rotation_apply(
 
 template<class ExecutionPolicy,
          class ElementType1,
-         ptrdiff_t ext1,
+         extents<>::size_type ext1,
          class Layout1,
          class Accessor1,
          class ElementType2,
-         ptrdiff_t ext2,
+         extents<>::size_type ext2,
          class Layout2,
          class Accessor2,
          std::floating_point Real>
@@ -390,11 +388,11 @@ void givens_rotation_apply(
 }
 
 template<class ElementType1,
-         ptrdiff_t ext1,
+         extents<>::size_type ext1,
          class Layout1,
          class Accessor1,
          class ElementType2,
-         ptrdiff_t ext2,
+         extents<>::size_type ext2,
          class Layout2,
          class Accessor2,
          std::floating_point Real>
@@ -405,7 +403,7 @@ void givens_rotation_apply(
   const complex<Real> s)
 {
   using std::conj;
-  for (ptrdiff_t i = 0; i < x.extent(0); ++i) {
+  for (extents<>::size_type i = 0; i < x.extent(0); ++i) {
     const auto dtemp = c * x(i) + s * y(i);
     y(i) = c * y(i) - conj(s) * x(i);
     x(i) = dtemp;
@@ -414,11 +412,11 @@ void givens_rotation_apply(
 
 template<class ExecutionPolicy,
          class ElementType1,
-         ptrdiff_t ext1,
+         extents<>::size_type ext1,
          class Layout1,
          class Accessor1,
          class ElementType2,
-         ptrdiff_t ext2,
+         extents<>::size_type ext2,
          class Layout2,
          class Accessor2,
          std::floating_point Real>
--- a/include/experimental/__p1673_bits/blas1_givens.hpp
+++ b/include/experimental/__p1673_bits/blas1_givens.hpp
@@ -344,12 +344,20 @@ label20:
   }
 }
 
-template<class inout_vector_1_t,
-         class inout_vector_2_t,
+// FIXME (Hoemmen 2021/05/28) Latest version of P0009 (mdspan) uses size_t
+// instead of ptrdiff_t, but the implementation hasn't changed yet.
+template<class ElementType1,
+         ptrdiff_t ext1,
+         class Layout1,
+         class Accessor1,
+         class ElementType2,
+         ptrdiff_t ext2,
+         class Layout2,
+         class Accessor2,
          class Real>
 void givens_rotation_apply(
-  inout_vector_1_t x,
-  inout_vector_2_t y,
+  std::experimental::basic_mdspan<ElementType1, std::experimental::extents<ext1>, Layout1, Accessor1> x,
+  std::experimental::basic_mdspan<ElementType2, std::experimental::extents<ext2>, Layout2, Accessor2> y,
   const Real c,
   const Real s)
 {
@@ -361,25 +369,37 @@ void givens_rotation_apply(
 }
 
 template<class ExecutionPolicy,
-         class inout_vector_1_t,
-         class inout_vector_2_t,
+         class ElementType1,
+         ptrdiff_t ext1,
+         class Layout1,
+         class Accessor1,
+         class ElementType2,
+         ptrdiff_t ext2,
+         class Layout2,
+         class Accessor2,
          class Real>
 void givens_rotation_apply(
   ExecutionPolicy&& /* exec */,
-  inout_vector_1_t x,
-  inout_vector_2_t y,
+  std::experimental::basic_mdspan<ElementType1, std::experimental::extents<ext1>, Layout1, Accessor1> x,
+  std::experimental::basic_mdspan<ElementType2, std::experimental::extents<ext2>, Layout2, Accessor2> y,
   const Real c,
   const Real s)
 {
   givens_rotation_apply(x, y, c, s);
 }
 
-template<class inout_vector_1_t,
-         class inout_vector_2_t,
+template<class ElementType1,
+         ptrdiff_t ext1,
+         class Layout1,
+         class Accessor1,
+         class ElementType2,
+         ptrdiff_t ext2,
+         class Layout2,
+         class Accessor2,
          class Real>
 void givens_rotation_apply(
-  inout_vector_1_t x,
-  inout_vector_2_t y,
+  std::experimental::basic_mdspan<ElementType1, std::experimental::extents<ext1>, Layout1, Accessor1> x,
+  std::experimental::basic_mdspan<ElementType2, std::experimental::extents<ext2>, Layout2, Accessor2> y,
   const Real c,
   const complex<Real> s)
 {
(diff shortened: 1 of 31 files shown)


include/experimental/__p1673_bits/blas1_matrix_frob_norm.hpp
```diff
@@ -51,10 +51,15 @@ namespace experimental {
 inline namespace __p1673_version_0 {
 namespace linalg {
 
-template<class in_matrix_t,
-         class Scalar>
-Scalar matrix_frob_norm(in_matrix_t A,
-                        Scalar init)
+template<
+    class ElementType,
+    class Extents,
```
mhoemmen, line 56: Instead of an `Extents` template parameter, would you consider instead adding `ptrdiff_t numRows` and `ptrdiff_t numCols` template parameters, and using `std::experimental::extents<numRows, numCols>` instead of `Extents` in the type of `A`?

mhoemmen (commented):
Please see comment -- thanks!

include/experimental/__p1673_bits/blas1_matrix_frob_norm.hpp
mhoemmen: This would make the matrix functions more consistent with the vector functions below.

amklinv-nnl: Will do!

include/experimental/__p1673_bits/blas1_givens.hpp
```diff
@@ -344,12 +344,20 @@ void givens_rotation_setup(const complex<Real>& f,
...
+         class Accessor1,
+         class ElementType2,
+         ptrdiff_t ext2,
+         class Layout2,
+         class Accessor2,
          class Real>
 void givens_rotation_apply(
-  inout_vector_1_t x,
-  inout_vector_2_t y,
+  std::experimental::basic_mdspan<ElementType1, std::experimental::extents<ext1>, Layout1, Accessor1> x,
```
mhoemmen, line 359: Excellent; thank you! : - D

include/experimental/__p1673_bits/blas1_givens.hpp
```diff
@@ -344,12 +344,20 @@ void givens_rotation_setup(const complex<Real>& f,
...
+  std::experimental::basic_mdspan<ElementType2, std::experimental::extents<ext2>, Layout2, Accessor2> y,
   const Real c,
```
mhoemmen, line 361: It occurs to me that the proposal constrains Givens rotations to work only on floating-point types, but we can always go back to fix that later.  (The relevant concept is `std::floating_point`, so we could say `std::floating_point Real` instead of `class Real`.)

amklinv-nnl: BRB, installing gcc 10 so I can actually `#include<concepts>`

include/experimental/__p1673_bits/blas1_linalg_add.hpp
```diff
@@ -97,13 +129,23 @@ void add(in_object_1_t x,
...
+         class Extents_y,
+         class Layout_y,
+         class Accessor_y,
+         class ElementType_z,
+         class Extents_z,
+         class Layout_z,
+         class Accessor_z>
+void add(
+  ExecutionPolicy&& /* exec */,
+  std::experimental::basic_mdspan<ElementType_x, Extents_x, Layout_x, Accessor_x> x,
```
mhoemmen, line 146: I'll need to think a bit how to constrain (not mandate -- mandating would be easier) x, y, and z to all have the same rank, but this is perfectly fine for now -- thank you! : - D

mhoemmen: I think we could do it like this:

1. Make the extents for each of x, y, and z a `ptrdiff_t...` parameter pack (e.g., `ptrdiff_t... Extents_x`)
2. Add a `requires` clause (establishing a constraint) after the template parameters: `requires sizeof...(Extents_x) == sizeof...(Extents_y) && sizeof...(Extents_x) == sizeof...(Extents_y)`

Daisy is crazy good at parameter pack magic.  I don't think we need it to be too fancy here.  Anyway, we'll need to make several passes to catch all the constraints, so it's totally up to you if we want to do this now or later.  Thanks! : - D

amklinv-nnl: I'll just do it now so I get better at constraints :-)

Thanks for the suggestion!

include/experimental/__p1673_bits/blas1_givens.hpp
```diff
@@ -190,7 +191,7 @@ void givens_rotation_setup(const Real f,
 }
 
 namespace impl {
-template<class Real>
+template<std::floating_point Real>
```
mhoemmen, line 194: This change is harmless, but [technically unnecessary](https://eel.is/c++draft/complex.numbers.general); instantiating `std::complex<T>` with `T` other than `float`, `double`, or `long double` has unspecified effect.  It's totally OK to leave this change, though, as it could make error messages nicer on some implementations.

mhoemmen (approved):
Thanks! : - D

mhoemmen:
The build weirdly can't find the `<concepts>` header:
```
 -- Detected support for C++20 standard
-- -- Using "-fconcepts" to enable concepts support
-- Build include directory: /home/runner/work/stdBLAS/stdBLAS/stdblas-build/include/experimental
-- Configuring done
-- Generating done
-- Build files have been written to: /home/runner/work/stdBLAS/stdBLAS/stdblas-build
[  2%] Building CXX object tests/CMakeFiles/matrix_inf_norm.dir/matrix_inf_norm.cpp.o
In file included from /home/runner/work/stdBLAS/stdBLAS/stdblas-src/include/experimental/linalg:56,
                 from /home/runner/work/stdBLAS/stdBLAS/stdblas-src/tests/matrix_inf_norm.cpp:1:
/home/runner/work/stdBLAS/stdBLAS/stdblas-src/include/experimental/__p1673_bits/blas1_givens.hpp:48:10: fatal error: concepts: No such file or directory
   48 | #include <concepts>
      |          ^~~~~~~~~~
```

amklinv-nnl:
> The build weirdly can't find the `<concepts>` header:

The CI build was using gcc 9, which does not support concepts. This was fixed in a separate PR which you accepted (thanks!)

include/experimental/__p1673_bits/blas1_vector_idx_abs_max.hpp
```diff
@@ -48,24 +48,23 @@ namespace experimental {
...
-ptrdiff_t vector_idx_abs_max(
+extents<>::size_type vector_idx_abs_max(
   std::experimental::basic_mdspan<ElementType, std::experimental::extents<ext0>, Layout, Accessor> v)
 {
   using std::abs;
+  using size_type = typename extents<>::size_type;
   using magnitude_type = decltype(abs(v(0)));
 
   if (v.extent(0) == 0) {
     return -1;
```
mhoemmen, line 63: Uh oh, we're returning -1 here.  This is actually a bug in the proposal itself (it says it returns -1 in this case).  I'll add an issue to fix the proposal.  We can come back around and fix this code later.

mhoemmen: Also, this function is called `idx_abs_max` in the proposal.  I'll file a separate issue for that too.

include/experimental/__p1673_bits/blas2_matrix_vector_product.hpp
```diff
@@ -350,9 +364,9 @@ void triangular_matrix_vector_product(in_matrix_t A,
     }
   }
   else {
-    for (ptrdiff_t j = 0; j < A.extent(1); ++j) {
-      const ptrdiff_t i_upper = explicitDiagonal ? j : j - ptrdiff_t(1);
-      for (ptrdiff_t i = 0; i <= i_upper; ++i) {
+    for (size_type j = 0; j < A.extent(1); ++j) {
+      const size_type i_upper = explicitDiagonal ? j : j - size_type(1);
```
mhoemmen, line 368: This will break on the first iteration, where `i_upper` will be `size_type(-1)`, i.e., `std::numeric_limits<size_type>::max()`.

include/experimental/__p1673_bits/blas2_matrix_vector_product.hpp
```diff
@@ -411,9 +427,9 @@ void triangular_matrix_vector_product(in_matrix_t A,
...
```
mhoemmen, line 431: Please see above note.

include/experimental/__p1673_bits/blas2_matrix_vector_solve.hpp
```diff
@@ -64,20 +64,21 @@ void trsv_upper_triangular_left_side(
...
   constexpr bool explicit_diagonal =
     std::is_same_v<DiagonalStorage, explicit_diagonal_t>;
+  using size_type = typename extents<>::size_type;
 
-  const ptrdiff_t A_num_rows = A.extent(0);
-  const ptrdiff_t A_num_cols = A.extent(1);
+  const size_type A_num_rows = A.extent(0);
+  const size_type A_num_cols = A.extent(1);
 
   // One advantage of using signed index types is that you can write
```
mhoemmen, line 72: Alas, we're not using signed index types any more, so this code will loop forever (`i >= 0`, always).

include/experimental/__p1673_bits/blas3_triangular_matrix_matrix_solve.hpp
```diff
@@ -62,21 +62,22 @@ void trsm_upper_triangular_left_side(
...
+  using size_type = typename extents<>::size_type;
 
-  const ptrdiff_t A_num_rows = A.extent(0);
-  const ptrdiff_t B_num_cols = B.extent(1);
+  const size_type A_num_rows = A.extent(0);
+  const size_type B_num_cols = B.extent(1);
 
-  for (ptrdiff_t k = 0; k < B_num_cols; ++k) {
+  for (size_type k = 0; k < B_num_cols; ++k) {
     // One advantage of using signed index types is that you can write
```
mhoemmen, line 71: Please see above note.

tests/CMakeLists.txt
```diff
@@ -21,7 +21,11 @@ linalg_add_test(dot)
 linalg_add_test(gemm)
 linalg_add_test(gemv)
 linalg_add_test(givens)
-linalg_add_test(iterator)
+# (AMK 6/7/21) Temporarily removing iterator test since it doesn't build
+# [ 73%] Building CXX object tests/CMakeFiles/iterator.dir/iterator.cpp.o
+# /home/amklinv/stdBLAS/tests/iterator.cpp:12:28: error: 'all' has not been declared in 'std::experimental'
```
mhoemmen, line 26: The `all` value is called `full_extent` now:

https://github.com/kokkos/mdspan/blob/stable/include/experimental/__p0009_bits/full_extent_t.hpp

We would have preferred `all`, but it would have been a bit of a bold name grab in the `std` namespace.

mhoemmen (changes requested):
Alas, I didn't think to warn you about the few places that would need fixing on changing from signed to unsigned index type.

mhoemmen (approved):

mhoemmen (approved):
Thanks! : - )

include/experimental/__p1673_bits/blas2_matrix_vector_product.hpp
```diff
@@ -350,9 +525,9 @@ void triangular_matrix_vector_product(in_matrix_t A,
     }
   }
   else {
-    for (ptrdiff_t j = 0; j < A.extent(1); ++j) {
-      const ptrdiff_t i_upper = explicitDiagonal ? j : j - ptrdiff_t(1);
-      for (ptrdiff_t i = 0; i <= i_upper; ++i) {
+    for (size_type j = 0; j < A.extent(1); ++j) {
+      const ptrdiff_t i_upper = explicitDiagonal ? j : j - size_type(1);
```
mhoemmen, line 529: The difference between two `size_t` is (unfortunately) still `size_t`, so this might wrap around.

https://godbolt.org/z/aMjMcKMEW

include/experimental/__p1673_bits/blas2_matrix_vector_product.hpp
```diff
@@ -411,9 +613,9 @@ void triangular_matrix_vector_product(in_matrix_t A,
...
```
mhoemmen, line 617: Please see above.

include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -611,54 +721,84 @@ void symmetric_matrix_product(
...
 {
+  using size_type = typename extents<>::size_type;
+
   if constexpr (std::is_same_v<Side, left_side_t>) {
     if constexpr (std::is_same_v<Triangle, lower_triangle_t>) {
-      for (ptrdiff_t j = 0; j < C.extent(1); ++j) {
-        for (ptrdiff_t i = j; i < C.extent(0); ++i) {
-          C(i,j) = typename out_matrix_t::value_type{};
-          for (ptrdiff_t k = 0; k < A.extent(1); ++k) {
+      for (size_type j = 0; j < C.extent(1); ++j) {
```
mhoemmen, line 788: Did I forget to account for implicit diagonal storage here?

mhoemmen (changes requested):
Thanks so much for this fix!  I just found the one issue.

mhoemmen:
Build failures are due to `basic_mdspan` removal and replacement by `mdspan`, which is no longer an alias.  I'll merge for now and we can do a global search and replace to fix the build errors afterwards.

PR #69, opened 2021-06-05
CI: Prepare for concepts

amklinv-nnl:
Github actions' default compiler is gcc 9. 10 is needed to support concepts. As a result, I made a custom docker container containing cmake, gcc 10, and googletest. Note that gcc11 generated compiler errors with the iterator code.


mhoemmen (approved):

mhoemmen:
@amklinv-nnl Thanks for doing this! : - D

Issue #66, opened 2021-05-28
P1673: Non-unique layouts?


mhoemmen:
1. Constrain provided algorithms only to take unique layouts?
2. Users can provide their own specializations for non-unique layouts?

mhoemmen:
Please see also https://github.com/kokkos/stdBLAS/issues/86 .

PR #82, opened 2021-06-18
Modify implementation to match proposal

amklinv-nnl:
Specifically dealing with constraints and mandates.
See #66.

### DIFF
--- a/include/experimental/__p1673_bits/blas1_linalg_swap.hpp
+++ b/include/experimental/__p1673_bits/blas1_linalg_swap.hpp
@@ -64,6 +64,10 @@ void swap_rank_1(
   std::experimental::mdspan<ElementType_x, std::experimental::extents<ext_x>, Layout_x, Accessor_x> x,
   std::experimental::mdspan<ElementType_y, std::experimental::extents<ext_y>, Layout_y, Accessor_y> y)
 {
+  static_assert(x.static_extent(0) == dynamic_extent ||
+                y.static_extent(0) == dynamic_extent ||
+                x.static_extent(0) == y.static_extent(0));
+
   using std::swap;
   using size_type = typename extents<>::size_type;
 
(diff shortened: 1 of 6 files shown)


include/experimental/__p1673_bits/blas1_linalg_swap.hpp
```diff
@@ -111,10 +122,10 @@ void swap_elements(
...
   std::experimental::mdspan<ElementType_y, std::experimental::extents<ext_y ...>, Layout_y, Accessor_y> y)
 {
-  if constexpr (x.rank() == 1) {
+  if constexpr (x.rank() == 1 && y.rank() == 1) {
     swap_rank_1(x, y);
   }
-  else if constexpr (x.rank() == 2) {
+  else if constexpr (x.rank() == 2 && y.rank() == 2) {
     swap_rank_2(x, y);
   }
```
mhoemmen, line 130: Did I write this `static_assert`?!?  It's definitely wrong.

1. `"Not implemented"` is `true` in a `bool` context, so `static_assert("Not implemented")` is just `static_assert(true)`.
2. One can't actually SFINAE out `static_assert(false)`.  Here is an example: https://godbolt.org/z/eY565W14n

One can google "dependent false" for a standard technique for dealing with (2).  However, we don't need it in this case; instead of an `else` clause, we could just have `else if constexpr (x.rank() > 2 or y.rank() > 2)`.

mhoemmen: I've filed this as a separate issue, https://github.com/kokkos/stdBLAS/issues/83 .

include/experimental/__p1673_bits/blas2_matrix_vector_product.hpp
```diff
@@ -66,6 +66,7 @@ template<class ElementType_A,
          extents<>::size_type ext_y,
          class Layout_y,
          class Accessor_y>
+  requires (Layout_A::is_always_unique())
```
mhoemmen, line 69: I think we discussed this one -- `Layout_A::template mapping<extents<numRows_A, numCols_A>>::is_always_unique()`.

mhoemmen (changes requested):

PR #84, opened 2021-06-20
Fetch dependencies from Github if not available

upsj:
Addressing #2, this PR fetches GoogleTest (and mdspan, as it currently relies on an installed mdspan, which causes a plain `cmake ..` to fail) as part of the configure step and builds them as part of the regular build step.

I needed to add two workarounds for GTest-specific issues on MSVC, which are
1. GTest seems to assume we have pthread support for some reason, so we need to explicitly disable it
2. GTest modifies `CMAKE_CXX_FLAGS*` with respect to `/MD` and `/MT`, so we need to force it to use the default `/MD`, which means we link against a shared CRT.

How could these changes impact the library?
If somebody included stdBLAS with tests enabled as a subdirectory in their own project, didn't have `GTest` installed and did build `GTest` themselves, these two variables would be overwritten. We could probably avoid that by checking whether the variables are set, or disabling tests by default if we are being built as part of another project.

I tested everything on GCC 11 and MSVC 19, so no obvious compilation issues :)

TODO: 
- [ ] Enable tests by default?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,7 +71,7 @@ if (NOT mdspan_FOUND)
   FetchContent_Declare(
     mdspan
     GIT_REPOSITORY https://github.com/kokkos/mdspan.git
-    GIT_TAG        5694f21c39f3b948d06a0c63b9c219bf802e28a8
+    GIT_TAG        stable
   )
   FetchContent_GetProperties(mdspan)
   if(NOT mdspan_POPULATED)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,7 +64,21 @@ endif()
 
 ################################################################################
 
-find_package(mdspan REQUIRED)
+find_package(mdspan QUIET)
+if (NOT mdspan_FOUND)
+  message(STATUS "No installed mdspan found, fetching from Github")
+  include(FetchContent)
+  FetchContent_Declare(
+    mdspan
+    GIT_REPOSITORY https://github.com/kokkos/mdspan.git
+    GIT_TAG        5694f21c39f3b948d06a0c63b9c219bf802e28a8
+  )
+  FetchContent_GetProperties(mdspan)
+  if(NOT mdspan_POPULATED)
+    FetchContent_Populate(mdspan)
+    add_subdirectory(${mdspan_SOURCE_DIR} ${mdspan_BINARY_DIR} EXCLUDE_FROM_ALL)
+  endif()
+endif()
 
 find_package(BLAS)
 option(LINALG_ENABLE_BLAS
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -64,7 +64,21 @@ endif()
...
 
-find_package(mdspan REQUIRED)
+find_package(mdspan QUIET)
+if (NOT mdspan_FOUND)
+  message(STATUS "No installed mdspan found, fetching from Github")
+  include(FetchContent)
+  FetchContent_Declare(
+    mdspan
+    GIT_REPOSITORY https://github.com/kokkos/mdspan.git
+    GIT_TAG        5694f21c39f3b948d06a0c63b9c219bf802e28a8
```
mhoemmen, line 74: The implementation of mdspan has been changing lately to track P0009 changes, so we tend to like fetching from `HEAD` of the `stable` branch.  Is there a way to do that, or do we have to pick a specific commit?

tests/CMakeLists.txt
```diff
@@ -1,4 +1,25 @@
-find_package(GTest REQUIRED)
+find_package(GTest)
+if (NOT GTest_FOUND)
+  message(STATUS "No installed GTest found, fetching from Github")
+  include(FetchContent)
+  FetchContent_Declare(
+    googletest
+    GIT_REPOSITORY https://github.com/google/googletest.git
+    GIT_TAG        7153098229e88295f9655ff1d3b0e2fa9eada5f8
```
mhoemmen, line 8: Is there a way to attach this to a particular release tag, rather than a SHA1?  Thanks! : - )

mhoemmen (commented):
Thanks so much for your contribution! : - ) I'd like @amklinv-nnl , who has been doing all of the CI automated testing work (and implementation work besides), to have a look before we approve this.

CMakeLists.txt
upsj: Yes, `GIT_TAG` can be either a commit hash, a branch name or a tag name. The [`ExternalProject`](https://cmake.org/cmake/help/latest/module/ExternalProject.html) documentation provides more details on the possible parameters

mhoemmen: @upsj Could you actually please change this to fetch from HEAD of the stable branch?  Thanks! : - )

tests/CMakeLists.txt
mhoemmen: I'll rephrase: Could you please change this to fetch from the current latest release?  Thanks! : - )

mhoemmen (approved):
Excellent, thanks for the changes and for explaining `GIT_TAG` : - ) !  I spoke with @amklinv-nnl offline, who approves of the PR with these changes.

PR #85, opened 2021-07-21
Remove static_assert(true)

### LINKED ISSUE #83: Fix static_assert(false)
https://github.com/kokkos/stdBLAS/blob/fbfa0a0652c9372256d2384cec437ef4129c3202/include/experimental/__p1673_bits/blas1_linalg_copy.hpp#L115

Did I write this `static_assert`?!?  It's definitely wrong.

1. `"Not implemented"` is `true` in a `bool` context, so `static_assert("Not implemented")` is just `static_assert(true)`.
2. One can't actually SFINAE out `static_assert(false)`.  Here is an example: https://godbolt.org/z/eY565W14n

One can google "dependent false" for a standard technique for dealing with (2).  However, we don't need it in this case; instead of an `else` clause, we could just have `else if constexpr (x.rank() > 2 or y.rank() > 2)`.

amklinv-nnl:
Closes #83

### DIFF
--- a/include/experimental/__p1673_bits/blas1_linalg_add.hpp
+++ b/include/experimental/__p1673_bits/blas1_linalg_add.hpp
@@ -151,14 +151,13 @@ void add(
   std::experimental::mdspan<ElementType_y, std::experimental::extents<ext_y ...>, Layout_y, Accessor_y> y,
   std::experimental::mdspan<ElementType_z, std::experimental::extents<ext_z ...>, Layout_z, Accessor_z> z)
 {
+  static_assert(z.rank() <= 2);
+  
   if constexpr (z.rank() == 1) {
     add_rank_1 (x, y, z);
   }
   else if constexpr (z.rank() == 2) {
     add_rank_2 (x, y, z);
-  }
-  else if constexpr (z.rank() > 2) {
-    static_assert(false);
   }
 }
 
--- a/include/experimental/__p1673_bits/blas1_linalg_add.hpp
+++ b/include/experimental/__p1673_bits/blas1_linalg_add.hpp
@@ -67,6 +67,16 @@ void add_rank_1(
   std::experimental::mdspan<ElementType_y, std::experimental::extents<ext_y>, Layout_y, Accessor_y> y,
   std::experimental::mdspan<ElementType_z, std::experimental::extents<ext_z>, Layout_z, Accessor_z> z)
 {
+  static_assert(x.static_extent(0) == dynamic_extent ||
+                z.static_extent(0) == dynamic_extent ||
+                x.static_extent(0) == z.static_extent(0));
+  static_assert(y.static_extent(0) == dynamic_extent ||
+                z.static_extent(0) == dynamic_extent ||
+                y.static_extent(0) == z.static_extent(0));
+  static_assert(x.static_extent(0) == dynamic_extent ||
+                y.static_extent(0) == dynamic_extent ||
+                x.static_extent(0) == y.static_extent(0));
+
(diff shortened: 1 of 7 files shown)


include/experimental/__p1673_bits/blas1_linalg_add.hpp
```diff
@@ -92,6 +102,26 @@ void add_rank_2(
   std::experimental::mdspan<ElementType_y, std::experimental::extents<numRows_y, numCols_y>, Layout_y, Accessor_y> y,
   std::experimental::mdspan<ElementType_z, std::experimental::extents<numRows_z, numCols_z>, Layout_z, Accessor_z> z)
 {
+  static_assert(x.static_extent(0) == dynamic_extent ||
```
mhoemmen, line 105: I do wish `constexpr` loops were a thing.  There are proposals....

include/experimental/__p1673_bits/blas1_linalg_add.hpp
```diff
@@ -121,15 +151,14 @@ void add(
   std::experimental::mdspan<ElementType_y, std::experimental::extents<ext_y ...>, Layout_y, Accessor_y> y,
   std::experimental::mdspan<ElementType_z, std::experimental::extents<ext_z ...>, Layout_z, Accessor_z> z)
 {
+  static_assert(z.rank() <= 2);
```
mhoemmen, line 154: This is harmless and please don't trouble yourself, but generally I think we want to specify the rank bound as a Constraint ("does not participate in overload resolution"), rather than a Mandate (`static_assert`).

include/experimental/__p1673_bits/blas1_linalg_copy.hpp
```diff
@@ -105,15 +116,14 @@ void copy(
   std::experimental::mdspan<ElementType_x, std::experimental::extents<ext_x ...>, Layout_x, Accessor_x> x,
   std::experimental::mdspan<ElementType_y, std::experimental::extents<ext_y ...>, Layout_y, Accessor_y> y)
 {
+  static_assert (x.rank() <= 2);
```
mhoemmen, line 119: Please see above note.

include/experimental/__p1673_bits/blas1_linalg_add.hpp
mhoemmen: Actually... don't worry about this for now, please : - ) .  I'll file an issue with the proposal itself, since it's not 100% clear to me whether it specifies this as a Constraint or Mandate.

include/experimental/__p1673_bits/blas2_matrix_vector_product.hpp
```diff
@@ -123,6 +124,7 @@ template<class ElementType_A,
          extents<>::size_type ext_z,
          class Layout_z,
          class Accessor_z>
+  requires (Layout_A::template mapping<std::experimental::extents<numRows_A, numCols_A>>::is_always_unique())
```
mhoemmen, line 127: This is not actually part of the formal wording, though it does appear once in the proposal.  The Requirements in **linalg.algs.reqs** only asks for "unique layout," which implies `is_unique()`, not necessarily `is_always_unique()`.  I'll file an issue with the proposal to make this more clear -- I think what we want is `is_unique()`, with burden on the caller to check this.

mhoemmen:
I'm inclined to merge this for now, then fix the proposal's wording.

mhoemmen:
Per discussion elsewhere, I'll merge for now, fixing the issue, then file issues with P1673's wording to address concerns about `is_unique` vs. `is_always_unique`.  Thanks @amklinv-nnl ! : - D

PR #87, opened 2021-07-22
Add triangular matrix functions and ExecutionPolicy overloads

### LINKED ISSUE #45: Add ExecutionPolicy overloads of in-place triangular_matrix_* functions (see list)
Add `ExecutionPolicy` overloads of the in-place versions of `triangular_matrix_vector_solve`, `triangular_matrix_left_product`, `triangular_matrix_right_product`, `triangular_matrix_matrix_left_solve`, and `triangular_matrix_matrix_right_solve`.

amklinv-nnl:
Closes #45


amklinv-nnl:
I added the overwriting triangular_matrix_[left/right]_product functions, but not the updating ones. Submitting a draft request in case @mhoemmen gets to it first.

include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -537,9 +537,577 @@ void matrix_product(
...
+         class Layout_B,
+         class Accessor_B,
+         class ElementType_C,
+         extents<>::size_type numRows_C,
+         extents<>::size_type numCols_C,
+         class Layout_C,
+         class Accessor_C>
+void triangular_matrix_left_product(
+  std::experimental::mdspan<ElementType_A, std::experimental::extents<numRows_A, numCols_A>, Layout_A, Accessor_A> A,
+  Triangle t,
```
mhoemmen, line 560: The `triangular_matrix_{left,right}_product` functions also take a `DiagonalStorage` template parameter and function parameter.  Please see e.g., Section **[linalg.algs.blas3.trmm.ov.left]**.   The two valid `DiagonalStorage` types are `implicit_unit_diagonal_t` and `explicit_diagonal_t`.

include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -537,9 +537,577 @@ void matrix_product(
...
+  std::experimental::mdspan<ElementType_B, std::experimental::extents<numRows_B, numCols_B>, Layout_B, Accessor_B> B,
+  std::experimental::mdspan<ElementType_C, std::experimental::extents<numRows_C, numCols_C>, Layout_C, Accessor_C> C)
+{
+  using size_type = typename extents<>::size_type;
+
+  if constexpr (std::is_same_v<Triangle, lower_triangle_t>) {
+    for (size_type j = 0; j < C.extent(1); ++j) {
+      for (size_type i = j; i < C.extent(0); ++i) {
```
mhoemmen, line 718: "Unlike the symmetric rank-1 or rank-k update functions, these functions (that is, `{symmetric,hermitian}_matrix_{left,right}_product`) assume that the input matrix `A` -- not the output matrix -- is symmetric."  Thus, they should iterate over all entries of C, but only over the lower or upper triangle of A.

include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -537,9 +537,577 @@ void matrix_product(
...
+         extents<>::size_type numRows_B,
+         extents<>::size_type numCols_B,
+         class Layout_B,
+         class Accessor_B,
+         class ElementType_C,
+         extents<>::size_type numRows_C,
+         extents<>::size_type numCols_C,
+         class Layout_C,
+         class Accessor_C>
+void symmetric_matrix_product(
```
mhoemmen, line 855: This `symmetric_matrix_product` function isn't in the current version of P1673.  It's OK to leave in there, but it doesn't have to be there.

include/experimental/__p1673_bits/blas3_triangular_matrix_matrix_solve.hpp
```diff
@@ -246,6 +246,128 @@ void trsm_lower_triangular_right_side(
...
+         class Layout_B,
+         class Accessor_B,
+         class ElementType_X,
+         extents<>::size_type numRows_X,
+         extents<>::size_type numCols_X,
+         class Layout_X,
+         class Accessor_X>
+void triangular_matrix_matrix_left_solve(
+  std::experimental::mdspan<ElementType_A, std::experimental::extents<numRows_A, numCols_A>, Layout_A, Accessor_A> A,
+  Triangle t,
```
mhoemmen, line 268: Would you consider commenting out `t` (as in `Triangle /* t */`) to prevent build warnings for `t` being unused?

mhoemmen (changes requested):
Thanks for the PR and for catching the missing functions! : - )  I caught just a couple issues that we should consider fixing before merging.

include/experimental/__p1673_bits/blas3_matrix_product.hpp
amklinv-nnl: Done!

include/experimental/__p1673_bits/blas3_matrix_product.hpp
amklinv-nnl: Done!

include/experimental/__p1673_bits/blas3_matrix_product.hpp
amklinv-nnl: Taking out the trash!

include/experimental/__p1673_bits/blas3_matrix_product.hpp
amklinv-nnl: Pretty sure I made the exact same mistake with the symmetric functions, so I should probably fix that too.

include/experimental/__p1673_bits/blas3_triangular_matrix_matrix_solve.hpp
amklinv-nnl: Sure would! How did you trigger those warnings? I didn't see them.

include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -558,28 +559,39 @@ template<class ElementType_A,
...
   using size_type = typename extents<>::size_type;
+  constexpr bool explicitDiagonal =
+    std::is_same_v<DiagonalStorage, explicit_diagonal_t>;
 
   if constexpr (std::is_same_v<Triangle, lower_triangle_t>) {
     for (size_type j = 0; j < C.extent(1); ++j) {
       for (size_type i = 0; i < C.extent(0); ++i) {
         C(i,j) = ElementType_C{};
-        for (size_type k = 0; k <= i; ++k) {
+        const size_type k_upper = explicitDiagonal ? i : i - size_type(1);
```
mhoemmen, line 574: `size_type` is unsigned, so  if `i` is zero and `explicitDiagonal` is `false`, then `i - size_type(i)` will be a big positive number.  I think the fix would be to change the bounds of the for loop over `i`.

include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -631,28 +646,39 @@ template<class ElementType_A,
...
         }
+        if constexpr (! explicitDiagonal) {
+          C(i,j) += /* 1 times */ B(i,j);
+        }
       }
     }
   }
   else { // upper_triangle_t
     for (size_type j = 0; j < C.extent(1); ++j) {
+      const size_type k_upper = explicitDiagonal ? j : j - size_type(1);
```
mhoemmen, line 673: Please see note above -- thanks!

include/experimental/__p1673_bits/blas3_matrix_product.hpp
amklinv-nnl: Done!

include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -715,6 +715,120 @@ void triangular_matrix_right_product(
...
+        }
+        if constexpr (explicitDiagonal) {
+          C(k,j) *= A(k,k);
+        }
+      }
+    }
+  }
+  else { // lower_triangle_t
+    for (size_type j=0; j < C.extent(1); ++j) {
+      for (ptrdiff_t k=C.extent(0)-1; k >= 0; --k) {
```
mhoemmen, line 754: This should be correct, though I tend to prefer `k = N; k > 0; --k)` count-down loops, as they are more robust to changes in the index type.

include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -715,6 +715,120 @@ void triangular_matrix_right_product(
...
+{
+  using size_type = typename extents<>::size_type;
+  constexpr bool explicitDiagonal =
+    std::is_same_v<DiagonalStorage, explicit_diagonal_t>;
+
+  if constexpr (std::is_same_v<Triangle, upper_triangle_t>) {
+    for (size_type j=0; j < C.extent(1); ++j) {
+      for (size_type k=0; k < C.extent(0); ++k) {
+        for (size_type i=0; i < k; ++i) {
+          C(i,j) += C(k,j)*A(i,k);
```
mhoemmen, line 744: Order of multiplication is actually significant, since we support element types like quaternions with noncommutative multiplication.  This means that the "left product" matrix A actually needs to go on the left for each element-times-element multiplication.

include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -715,6 +715,120 @@ void triangular_matrix_right_product(
...
+        }
+        if constexpr (explicitDiagonal) {
+          C(k,j) *= A(k,k);
```
mhoemmen, line 747: Given that multiplication order is significant (see above), would you consider spelling out the multiplication instead of using `*=`, as a way to make clear the order of the factors?

include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -715,6 +715,120 @@ void triangular_matrix_right_product(
...
+         extents<>::size_type numRows_C,
+         extents<>::size_type numCols_C,
+         class Layout_C,
+         class Accessor_C>
+void triangular_matrix_right_product(
+  std::experimental::mdspan<ElementType_A, std::experimental::extents<numRows_A, numCols_A>, Layout_A, Accessor_A> A,
+  Triangle /* t */,
+  DiagonalStorage /* d */,
+  std::experimental::mdspan<ElementType_C, std::experimental::extents<numRows_C, numCols_C>, Layout_C, Accessor_C> C)
+{
```
mhoemmen, line 806: It looks like the implementation of this overload is missing.

tests/gemm.cpp
```diff
@@ -369,4 +370,394 @@ TEST(BLAS3_trmm, left_lower_tri_implicit_diag)
...
+      }
+    }
+  }
+
+TEST(BLAS3_trmm, right_lower_tri_implicit_diag)
+  {
+    /* C = A * B, where A is triangular mxm */
+    using extents_t = extents<dynamic_extent, dynamic_extent>;
+    using matrix_t = mdspan<double, extents_t, layout_left>;
+    double snan = std::numeric_limits<double>::signaling_NaN();
```
mhoemmen, line 573: Would you consider making this `constexpr`, given that it's a kind of named constant?

tests/gemm.cpp
```diff
@@ -369,4 +370,394 @@ TEST(BLAS3_trmm, left_lower_tri_implicit_diag)
...
+    gs(1,0) = -5.16;
+    gs(0,1) = 6.7;
+    gs(1,1) = 10.6;
+    gs(0,2) = 1.8;
+    gs(1,2) = 1.6;
+
+    for (ptrdiff_t j = 0; j < m; ++j) {
+      for (ptrdiff_t i = 0; i < n; ++i) {
+        // FIXME: Choose a more reasonable value for the tolerance
+        double tol = 1e-9;
```
mhoemmen, line 612: Would you consider making constants like these `constexpr` if possible?  Thanks!

tests/gemm.cpp
```diff
@@ -369,4 +370,394 @@ TEST(BLAS3_trmm, left_lower_tri_implicit_diag)
     }
   }
 
+TEST(BLAS3_trmm, left_upper_tri_explicit_diag)
```
mhoemmen, line 373: I do appreciate all the new tests -- thanks! : - D

include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -673,7 +673,7 @@ void triangular_matrix_right_product(
       const ptrdiff_t k_upper = explicitDiagonal ? j : j - size_type(1);
```
mhoemmen: The variable `j` has type `size_type` (`std::size_t`, an unsigned integer), and `j - size_type(1)` has type `size_type` (please see [this godbolt example](https://godbolt.org/z/fe6YEfPb3)), so `k_upper` will be the biggest possible 64-bit positive integer.  The resulting `k <= k_upper` comparison will always evaluate to `true`.

The work-around that comes to mind would be to have a special case for `explicitDiagonal` being `true`.  That way, one could rewrite the loop bound as `k < j`.  Here is a possible reimplementation of lines 676-681:
```
for (ptrdiff_t k = 0; k < j; ++k) {
  C(i,j) += B(i,k) * A(k,j);
}
if constexpr (explicitDiagonal) {
  C(i,j) += B(i,j) /* times 1 */;
} else {
  C(i,j) += B(i,j) * A(j,j);
}
```

mhoemmen (changes requested):

include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -572,7 +572,7 @@ void triangular_matrix_left_product(
       for (size_type i = 0; i < C.extent(0); ++i) {
         C(i,j) = ElementType_C{};
         const ptrdiff_t k_upper = explicitDiagonal ? i : i - size_type(1);
```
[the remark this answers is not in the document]
amklinv-nnl: Replaced with ptrdiff_t(1)

include/experimental/__p1673_bits/blas3_matrix_product.hpp
amklinv-nnl: Done!

include/experimental/__p1673_bits/blas3_matrix_product.hpp
amklinv-nnl: Done!

include/experimental/__p1673_bits/blas3_matrix_product.hpp
amklinv-nnl: Done!

include/experimental/__p1673_bits/blas3_matrix_product.hpp
amklinv-nnl: It sure was at the time! It should be there now.

include/experimental/__p1673_bits/blas3_triangular_matrix_matrix_solve.hpp
mhoemmen: @amklinv-nnl Greetings!  Sorry I missed your comment!  Adding `-Wall` to the list of warnings should help.

include/experimental/__p1673_bits/blas3_matrix_product.hpp
amklinv-nnl: Is changing size_type(1) to ptrdiff_t(1) acceptable? If not, I can change it.

include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -664,56 +1032,116 @@ template<class ElementType_A,
...
-          for (size_type k = 0; k < A.extent(1); ++k) {
-            C(i,j) += B(i,k) * A(k,j);
-          }
+  }
+  else { // upper_triangle_t
+    for (size_type j = 0; j < C.extent(1); ++j) {
+      for (size_type i = 0; i <= j; ++i) {
+        C(i,j) = E(i,j);
+        for (size_type k = 0; k < A.extent(1); ++k) {
+          C(i,j) += B(i,k) * A(k,j);
```
mhoemmen, line 1144: Both the symmetric and Hermitian left and right products should go over the matrix `A` twice, unlike the triangular left and right products.

mhoemmen (changes requested):

tests/gemm.cpp
amklinv-nnl: Done!

tests/gemm.cpp
amklinv-nnl: Done!

tests/gemm.cpp
amklinv-nnl: Untested code is broken code :-)

include/experimental/__p1673_bits/blas3_matrix_product.hpp
amklinv-nnl: I think we should separate that into its own issue.

mhoemmen: @amklinv-nnl I might be a bit confused perhaps -- were those functions implemented before?  If not and if this PR provides a partial, known-incorrect implementation for later revision, then would you consider putting `assert(false);` in the functions known to be correct, so that we know to come back to them later?  Thanks! : - )

include/experimental/__p1673_bits/blas3_matrix_product.hpp
mhoemmen: @amklinv-nnl Please see suggested changes below -- thanks!

include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -40,6 +40,8 @@
 //@HEADER
 */
 
+#include <cassert>
```
mhoemmen, line 43: We probably should put this inside the include-once guard, but no biggie : - )

include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -1206,23 +1170,26 @@ void hermitian_matrix_left_product(
...
+  using std::conj;
 
   if constexpr (std::is_same_v<Triangle, lower_triangle_t>) {
     for (size_type j = 0; j < C.extent(1); ++j) {
-      for (size_type i = j; i < C.extent(0); ++i) {
+      for (size_type i = 0; i < C.extent(0); ++i) {
         C(i,j) = ElementType_C{};
         for (size_type k = 0; k < A.extent(1); ++k) {
-          C(i,j) += A(i,k) * B(k,j);
+          ElementType_A aik = i <= k ? conj(A(k,i)) : A(i,k);
```
mhoemmen, line 1180: I don't think this will compile for generic `ElementType_A`, because `std::conj` always returns `std::complex` (even if its argument is real).  There might be something in here like a "conditional `conj`" that returns the same type as its argument, but it's not too hard to write one:

```cpp
template<class T>
T cond_conj(const T& t) {
  return t;
}
template<class R>
std::complex<R> cond_conj(const std::complex<R>& z)
{
  using std::conj;
  return conj(z); 
}
```

include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -1279,23 +1246,26 @@ void hermitian_matrix_right_product(
...
+  using std::conj;
 
   if constexpr (std::is_same_v<Triangle, lower_triangle_t>) {
     for (size_type j = 0; j < C.extent(1); ++j) {
-      for (size_type i = j; i < C.extent(0); ++i) {
+      for (size_type i = 0; i < C.extent(0); ++i) {
         C(i,j) = ElementType_C{};
         for (size_type k = 0; k < A.extent(1); ++k) {
-          C(i,j) += B(i,k) * A(k,j);
+          ElementType_A akj = j <= k ? A(k,j) : conj(A(j,k));
```
mhoemmen, line 1256: Please see note above; thanks!

Issue #100, opened 2021-10-11
`dotc(x,y)`: why is conjugate applied to second argument?


fnrizzi:
@mhoemmen 

I see that `dotc(x,y)` is doing: `dot(x, conjugate(y))`. 
But the BLAS `dotc` does: `dot(conjugate(x), y)`. 

is there a reason for the difference? Or am i totally missing something obvious here - which could very well be these days :) ?

mhoemmen:
@lums658 wrote:
> ... hope you don’t mind a lurker speaking up.

You're not just any lurker! ; - ) Please do feel welcome to speak up!  Thanks also for checking MTL4 and the literature!

We intend to do what the BLAS does as much as C++ interface idioms permit, so that vendors could just dispatch to the BLAS (if the types all match).

@fnrizzi wrote:
> I am happy to change the impl if you want

Please feel free to to do, and to submit a P1673 PR [here](https://github.com/ORNL/cpp-proposals-pub) if you have time -- thanks!

fnrizzi:
1. code PR ready: https://github.com/kokkos/stdBLAS/pull/103

2. > Please feel free to to do, and to submit a P1673 PR [here](https://github.com/ORNL/cpp-proposals-pub) 

@mhoemmen I edited the spec document, but I cannot push a branch to it, permission denied :)

PR #103, opened 2021-10-12
dotc: conjugate first argument, conforming to BLAS

fnrizzi:
related to #100

### DIFF
--- a/include/experimental/__p1673_bits/blas1_dot.hpp
+++ b/include/experimental/__p1673_bits/blas1_dot.hpp
@@ -50,6 +50,7 @@ namespace experimental {
 inline namespace __p1673_version_0 {
 namespace linalg {
 
+// dot: no policy, with init value
 template<class ElementType1,
          extents<>::size_type ext1,
          class Layout1,
@@ -73,6 +74,7 @@ Scalar dot(std::experimental::mdspan<ElementType1, std::experimental::extents<ex
   return init;
 }
 
+// dot: with policy, with init value
 template<class ExecutionPolicy,
          class ElementType1,
          extents<>::size_type ext1,
@@ -92,8 +94,7 @@ Scalar dot(
   return dot(v1, v2, init);
 }
 
-// Conjugated dot
-
+// conjugated dot: no policy, with init value
 template<class ElementType1,
          extents<>::size_type ext1,
          class Layout1,
@@ -111,6 +112,7 @@ Scalar dotc(
   return dot(conjugated(v1), v2, init);
 }
 
+// conjugated dot: with policy, with init value
 template<class ExecutionPolicy,
          class ElementType1,
          extents<>::size_type ext1,
@@ -148,22 +150,9 @@ namespace dot_detail {
     std::experimental::mdspan<ElementType1, std::experimental::extents<ext1>, Layout1, Accessor1> x,
     std::experimental::mdspan<ElementType2, std::experimental::extents<ext2>, Layout2, Accessor2> y)
   -> decltype(x(0) * y(0));
-
-  template<
-    class ElementType1,
-    extents<>::size_type ext1,
-    class Layout1,
-    class Accessor1,
-    class ElementType2,
-    extents<>::size_type ext2,
-    class Layout2,
-    class Accessor2>
-  auto dotc_return_type_deducer(
-    std::experimental::mdspan<ElementType1, std::experimental::extents<ext1>, Layout1, Accessor1> x,
-    std::experimental::mdspan<ElementType2, std::experimental::extents<ext2>, Layout2, Accessor2> y)
-    -> decltype(conj(x(0)) * y(0));
 } // namespace dot_detail
 
+// dot: no policy, no init value
 template<class ElementType1,
          extents<>::size_type ext1,
          class Layout1,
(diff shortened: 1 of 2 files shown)


include/experimental/__p1673_bits/blas1_dot.hpp
```diff
@@ -178,9 +192,9 @@ template<class ElementType1,
 auto dotc(
   std::experimental::mdspan<ElementType1, std::experimental::extents<ext1>, Layout1, Accessor1> v1,
   std::experimental::mdspan<ElementType2, std::experimental::extents<ext2>, Layout2, Accessor2> v2)
--> decltype(dot_detail::dot_return_type_deducer(v1, v2))
+-> decltype(dot_detail::dotc_return_type_deducer(v1, v2))
 {
-  using return_t = decltype(dot_detail::dot_return_type_deducer(v1, v2));
+  using return_t = decltype(dot_detail::dotc_return_type_deducer(v1, v2));
```
mhoemmen, line 197: Would you consider removing `dotc_return_type_deducer` and just having `dotc` return `auto`, as before?  Plain old `auto dotc` without an explicit return type should work just fine, because `dotc(x, y)` is just `dot(conjugated(x), y)`, and even C++14 knows how to deduce the function's return type from the `return` expression.

tests/dot.cpp
```diff
@@ -134,6 +134,58 @@ namespace {
...
     //EXPECT_EQ( dotResultPar, expectedDotResult );
   }
+
+  TEST(BLAS1_dot, mdspan_complex_double_test2)
+  {
+    using real_t = double;
+    using scalar_t = std::complex<real_t>;
+    using vector_t = mdspan<scalar_t, extents<dynamic_extent>>;
+
+    constexpr ptrdiff_t vectorSize(5);
```
mhoemmen, line 144: Would you consider using `size_t` instead of `ptrdiff_t`, given that `mdspan` has changed to use `size_t` for extents?

tests/dot.cpp
```diff
@@ -134,6 +134,58 @@ namespace {
     //dot (fake_executor_t (), x, y, dotResultPar);
     //EXPECT_EQ( dotResultPar, expectedDotResult );
   }
+
+  TEST(BLAS1_dot, mdspan_complex_double_test2)
```
mhoemmen, line 138: Thank you for fixing the test too! : - )

include/experimental/__p1673_bits/blas1_dot.hpp
```diff
@@ -108,7 +108,7 @@ Scalar dotc(
   std::experimental::mdspan<ElementType2, std::experimental::extents<ext2>, Layout2, Accessor2> v2,
   Scalar init)
 {
-  return dot(v1, conjugated(v2), init);
+  return dot(conjugated(v1), v2, init);
```
mhoemmen, line 111: Thanks for fixing this! : - D

include/experimental/__p1673_bits/blas1_dot.hpp
```diff
@@ -148,6 +148,20 @@ namespace dot_detail {
     std::experimental::mdspan<ElementType1, std::experimental::extents<ext1>, Layout1, Accessor1> x,
     std::experimental::mdspan<ElementType2, std::experimental::extents<ext2>, Layout2, Accessor2> y)
   -> decltype(x(0) * y(0));
+
+  template<
```
mhoemmen, line 152: Please see comment below -- thanks!

mhoemmen (changes requested):

include/experimental/__p1673_bits/blas1_dot.hpp
fnrizzi: you mean above about removing the dotc_return type?

include/experimental/__p1673_bits/blas1_dot.hpp
fnrizzi: @mhoemmen  
Just to clarify, before my changes, `dotc` looked like this:
```cpp
auto dotc(
  std::experimental::mdspan<ElementType1, std::experimental::extents<ext1>, Layout1, Accessor1> v1,
  std::experimental::mdspan<ElementType2, std::experimental::extents<ext2>, Layout2, Accessor2> v2)
-> decltype(dot_detail::dot_return_type_deducer(v1, v2))
```
so note that it was using `dot_return_type` , but WITHOUT saying `_deducer(conjugate(v1), v2)`. 
When I read the spec, I noticed this:
```
Effects: If in_vector_2_t::element_type is complex<R> for some R, let T be decltype(v1(0)*conj(v2(0))); 
else, let T be decltype(v1(0)*v2(0)). Then, the two-parameter overload is equivalent 
to dotc(v1, v2, T{});, and the three-parameter overload is equivalent to dotc(exec, v1, v2, T{});.
``` 
so that is why I added `dotc_return_type` with that syntax, to be consistent with what was written in the spec. 

So, assuming you still want to get rid of `dotc_return_type` do you want to leave:
```cpp
auto dotc(
  std::experimental::mdspan<ElementType1, std::experimental::extents<ext1>, Layout1, Accessor1> v1,
  std::experimental::mdspan<ElementType2, std::experimental::extents<ext2>, Layout2, Accessor2> v2)
-> decltype(dot_detail::dot_return_type_deducer(v1, v2))
{}
```
or should I just do:
```cpp
auto dotc(
  std::experimental::mdspan<ElementType1, std::experimental::extents<ext1>, Layout1, Accessor1> v1,
  std::experimental::mdspan<ElementType2, std::experimental::extents<ext2>, Layout2, Accessor2> v2)
  {}
```

mhoemmen: The latter two choices would be true to the proposal.  I would recommend the last choice: 
```cpp
auto dotc(
  std::experimental::mdspan<ElementType1, std::experimental::extents<ext1>, Layout1, Accessor1> v1,
  std::experimental::mdspan<ElementType2, std::experimental::extents<ext2>, Layout2, Accessor2> v2)
-> decltype(dot_detail::dot_return_type_deducer(conjugate(v1), v2))
```

tests/dot.cpp
fnrizzi: happy to help out: https://github.com/kokkos/stdBLAS/pull/104

include/experimental/__p1673_bits/blas1_dot.hpp
```diff
@@ -50,6 +50,7 @@ namespace experimental {
 inline namespace __p1673_version_0 {
 namespace linalg {
 
+// dot: no policy, with init value
```
mhoemmen, line 53: These comments are harmless, but generally unnecessary.  (The policy argument's template parameter is called `ExecutionPolicy`, and the init value is called `init`.)

fnrizzi: yes, it was more of a "description" rather than the exact wording, just my personal style, which I think it helps tracking visually code organization. Sorry for adding them without asking! Won't add from now on ;)

fnrizzi: It actually comes subconsciously

PR #104, opened 2021-10-13
tests: replace ptrdiff_t with std::size_t

fnrizzi:
- I replaced in all tests. 
- in the impl, I noticed that in some cases `ptrdiff_t` is used for valid reasons, so I have not touched the impl stuff


mhoemmen (approved):
Looks good -- thanks for fixing all these! : - D

PR #119, opened 2021-11-01
blas1 customization: address some PR comments

fnrizzi:
- fixed casting to use `static_cast<>()`
- change to use `is_same_v`
- fixed customization struct for non-void return functions to work as:
```cpp
template <class Exec, class v1_t, class v2_t, class Scalar>
struct is_custom_dot_avail<
  Exec, v1_t, v2_t, Scalar, std::void_t< ... >
  >
{
  static constexpr bool has_matching_return_type = std::is_same_v<
    Scalar, decltype(dot(...)) >;

  static constexpr bool is_not_inline_exec_tag = !std::is_same_v<
    Exec,std::experimental::linalg::impl::inline_exec_t>;

  static constexpr bool value = has_matching_return_type && is_not_inline_exec_tag;
};
```

### DIFF
--- a/examples/kokkos-based/add_kokkos.cpp
+++ b/examples/kokkos-based/add_kokkos.cpp
@@ -39,7 +39,7 @@ int main(int argc, char* argv[])
     std::vector<value_type> gold(N);
     for(std::size_t i=0; i<x.extent(0); i++){
       x(i) = i;
-      y(i) = i + (value_type)10;
+      y(i) = i + static_cast<value_type>(10);
       z(i) = 0;
       gold[i] = x(i) + y(i);
     }
--- a/include/experimental/__p1673_bits/blas1_dot.hpp
+++ b/include/experimental/__p1673_bits/blas1_dot.hpp
@@ -59,21 +59,29 @@ struct is_custom_dot_avail : std::false_type {};
 template <class Exec, class v1_t, class v2_t, class Scalar>
 struct is_custom_dot_avail<
   Exec, v1_t, v2_t, Scalar,
-  std::enable_if_t<
-    std::is_same<
-      decltype(dot
-	       (std::declval<Exec>(),
-		std::declval<v1_t>(),
-		std::declval<v2_t>(),
-		std::declval<Scalar>()
-		)
-	       ),
-      Scalar
-      >::value
+  std::void_t<
+    decltype(dot
+	     (std::declval<Exec>(),
+	      std::declval<v1_t>(),
+	      std::declval<v2_t>(),
+	      std::declval<Scalar>()
+	      )
+	     )
     >
   >
 {
-  static constexpr bool value = !std::is_same<Exec,std::experimental::linalg::impl::inline_exec_t>::value;
+  static constexpr bool has_matching_return_type = std::is_same_v<
+    Scalar,
+    decltype(
+	     dot(std::declval<Exec>(), std::declval<v1_t>(),
+		 std::declval<v2_t>(), std::declval<Scalar>())
+	     )
+    >;
+
+  static constexpr bool is_not_inline_exec_tag = !std::is_same_v<
+    Exec,std::experimental::linalg::impl::inline_exec_t>;
+
+  static constexpr bool value = has_matching_return_type && is_not_inline_exec_tag;
 };
 
 } // end anonymous namespace
--- a/include/experimental/__p1673_bits/blas1_vector_idx_abs_max.hpp
+++ b/include/experimental/__p1673_bits/blas1_vector_idx_abs_max.hpp
@@ -57,16 +57,21 @@ struct is_custom_idx_abs_max_avail : std::false_type {};
 template <class Exec, class v_t>
 struct is_custom_idx_abs_max_avail<
   Exec, v_t,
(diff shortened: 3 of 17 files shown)


examples/kokkos-based/add_kokkos.cpp
```diff
@@ -39,7 +39,7 @@ int main(int argc, char* argv[])
     std::vector<value_type> gold(N);
     for(std::size_t i=0; i<x.extent(0); i++){
       x(i) = i;
-      y(i) = i + (value_type)10;
+      y(i) = i + static_cast<value_type>(10);
```
mhoemmen, line 42: Thanks for making these changes! : - )

include/experimental/__p1673_bits/blas1_dot.hpp
```diff
@@ -59,21 +59,29 @@ struct is_custom_dot_avail : std::false_type {};
...
-      decltype(dot
-	       (std::declval<Exec>(),
-		std::declval<v1_t>(),
-		std::declval<v2_t>(),
-		std::declval<Scalar>()
-		)
-	       ),
-      Scalar
-      >::value
+  std::void_t<
```
mhoemmen, line 62: I'm not quite sure what you're trying to do here with `void_t`.  Are you trying to SFINAE out the body of this `struct` if `dot(Exec, v1_t, v2_1, Scalar)` is not well formed, or is this a redundant expression of `has_matching_return_type`?

include/experimental/__p1673_bits/blas1_dot.hpp
```diff
@@ -59,21 +59,29 @@ struct is_custom_dot_avail : std::false_type {};
...
+	      std::declval<v1_t>(),
+	      std::declval<v2_t>(),
+	      std::declval<Scalar>()
+	      )
+	     )
     >
   >
 {
-  static constexpr bool value = !std::is_same<Exec,std::experimental::linalg::impl::inline_exec_t>::value;
+  static constexpr bool has_matching_return_type = std::is_same_v<
```
mhoemmen, line 73: Would you consider making `has_matching_return_type` and `is_not_inline_exec_tag` both `private`, as they are implementation details of `value`?

include/experimental/__p1673_bits/blas1_vector_idx_abs_max.hpp
```diff
@@ -114,7 +124,7 @@ typename extents<>::size_type idx_abs_max(
   std::experimental::mdspan<ElementType, std::experimental::extents<ext0>, Layout, Accessor> v)
 {
 
-  // if vector is empty, always retun according to our proposal
+  // if vector is empty, always return according to our proposal
```
mhoemmen, line 127: Nice catch! : - )

mhoemmen (commented):
I'm not quite sure what you're trying to do with `void_t`.  It looks a bit like you're imitating the first example [on cppreference](https://en.cppreference.com/w/cpp/types/void_t), in order to check that a particular expression is well formed.  Is that the idea?

fnrizzi:
@mhoemmen re the `void_t` thing: this is based on your previous PR comment https://github.com/kokkos/stdBLAS/pull/101#discussion_r737626036

From that, I understood that you wanted both conditions to be in `value`.
Did I misunderstand it?

include/experimental/__p1673_bits/blas1_vector_idx_abs_max.hpp
fnrizzi: you pointed it out in the other PR :) so good catch you!

include/experimental/__p1673_bits/blas1_dot.hpp
fnrizzi: yes exactly, if this is not well formed it does not work. But if it, then I am also checking the return type is sam as `Scalar`. I was trying to address  your comment here: https://github.com/kokkos/stdBLAS/pull/101#discussion_r737626036

mhoemmen: Thanks @fnrizzi ! : - )  I think I see what the idea is here -- the `void_t` thing checks that the expression is well formed, and the `value` thing back-ports to C++ < 20 (else you would use `same_as` to constrain, then inherit from `true_type`).

One potential issue is that not all specializations of the struct inherit from either `true_type` or `false_type`.  However, I don't think that's a big deal for an implementation detail, as long as developers know to check `value` and not try to check inheritance.

mhoemmen:
@fnrizzi I think I see what the idea is here -- the `void_t` thing checks that the expression is well formed, and the `value` thing back-ports to C++ < 20 (else you would use `same_as` to constrain, then inherit from `true_type`).

One potential issue is that not all specializations of the struct inherit from either `true_type` or `false_type`.  However, I don't think that's a big deal for an implementation detail, as long as developers know to check `value` and not try to check inheritance.

We might like to fix inheritance from `true_type` / `false_type` at some point, and also make the implementation details in the `struct`s other than `value` `private`, but I'm happy to merge this as-is.  Thanks for your hard work!

fnrizzi:
yes, you got my idea! I will fix the private thing later!
thanks!

Issue #105, opened 2021-10-18
BLAS 3: missing `symmetric_matrix_rank_k_update` implementation


fnrizzi:
@mhoemmen 
this function is mentioned in the proposal, but it is missing, am I right?

PR #127, opened 2021-11-22
[ready] Blas3 matrix rank k update

fnrizzi:
This addresses #105


fnrizzi:
ready for review

mhoemmen (approved):
Thanks for the new code!  Oof, it looks like we didn't implement these functions at all before.  Could you please file a follow-on issue to add tests for these new functions?

fnrizzi:
> Thanks for the new code! Oof, it looks like we didn't implement these functions at all before. Could you please file a follow-on issue to add tests for these new functions?

done! Yes, I opened issue #105 a while back :)

include/experimental/__p1673_bits/blas3_matrix_rank_k_update.hpp
```diff
@@ -0,0 +1,473 @@
...
+  using std::conj;
+  constexpr bool lower_tri = std::is_same_v<Triangle, lower_triangle_t>;
+  using size_type = typename extents<>::size_type;
+
+  for (size_type j = 0; j < C.extent(1); ++j) {
+    const size_type i_lower = lower_tri ? j : size_type(0);
+    const size_type i_upper = lower_tri ? C.extent(0) : j+1;
+    for (size_type i = i_lower; i < i_upper; ++i) {
+      for (size_type k = 0; k < A.extent(1); ++k) {
+        C(i,j) += alpha * A(i,k) * conj(A(j,k));
```
mhoemmen, line 327: Sorry I missed this in my earlier review!  `conj` always returns `std::complex`, even the overloads that take a non-`complex` argument.  This could cause build errors.

A straightforward fix for now would be to use a generic lambda.  We'll generally need to do this wherever we use `conj`, unless we know that the value type is `complex`.
```
constexpr bool A_is_complex =
  std::is_same_v<typename decltype(A)::value_type, std::complex<float>> ||
  std::is_same_v<typename decltype(A)::value_type, std::complex<double>> ||
  std::is_same_v<typename decltype(A)::value_type, std::complex<long double>>;
auto conj_if_needed = [](const auto A_jk) { 
  if constexpr (A_is_complex) {
    return conj(A_jk);
  } else {
    return A_jk;
  }
};
```

mhoemmen (changes requested):

include/experimental/__p1673_bits/blas3_matrix_rank_k_update.hpp
fnrizzi: ha! Thanks! 
Ok, so you mean this needs to be done *everywhere* we use `conj`? i will hunt those down in a separate PR maybe

fnrizzi: @mhoemmen btw, enjoy thanksgiving :)

fnrizzi:
@mhoemmen I incorporated your suggestion with a slight modification , and used it for `rank_k_` and `rank_2k_`. 
Let me know what you think - if you approve, I will make changes to other instances where `conj` is used.

include/experimental/__p1673_bits/blas3_matrix_rank_k_update.hpp
mhoemmen: @fnrizzi wrote:
> For instance `rank_2k` uses that, in fact I used that as starting point :)

No promise that existing code was correct ; - )

There are two issues with using `conj` "generically" (for any element type):

1. It always returns `std::complex<U>` for some `U`.  The resulting arithmetic expression may not compile.
2. It only has overloads for floating-point and integer types.  Thus, it won't compile for types that are not implicitly convertible to one of these.

fnrizzi:
@mhoemmen I assume the thumbs up means you approve, so I will apply the change to every instance of `conj` that applies

fnrizzi:
@mhoemmen 
I made the changes to `conj` as follows:
- `./__p1673_bits/blas1_givens.hp`: here I did not change it because, where used, arguments are explicitly complex

- I changed it in all these:
```
./__p1673_bits/blas3_matrix_product.hpp
./__p1673_bits/blas2_matrix_rank_1_update.hpp
./__p1673_bits/blas2_matrix_rank_2_update.hpp
./__p1673_bits/blas2_matrix_vector_product.hpp
```

mhoemmen:
> I also had to make a small change to `conj_if_needed` since gcc10 was giving me headaches with const

True: `is_complex_v<decltype(value)>` should instead be `is_complex_v<std::remove_cv_t<decltype(value)>>`.

Thanks for your changes!  Is this PR still a work in progress, or is it ready to merge?

fnrizzi:
@mhoemmen ready to merge if you agree with all the changes! - sorry, I regularly forget to change the WIP :)

crtrott (approved):

include/experimental/__p1673_bits/conjugate_if_needed.hpp
```diff
@@ -0,0 +1,85 @@
...
+namespace std {
+namespace experimental {
+inline namespace __p1673_version_0 {
+namespace linalg {
+namespace impl{
+
+template<class T> struct is_complex : std::false_type{};
+
+template<> struct is_complex<std::complex<float>> : std::true_type{};
+template<> struct is_complex<std::complex<const float>> : std::true_type{};
```
mhoemmen, line 57: Oof, I missed this bit.  I don't think you can instantiate `std::complex` with a `const`-qualified type.  Did you perhaps mean to write `is_complex<const std::complex<float>>`?

I think a better strategy would be to define `is_complex` to inherit from `true_type` only for `std::complex<float>`, `std::complex<double>`, and `std::complex<long double>`.  Then, use `std::remove_cv_t` on types before applying `is_complex` to them.

include/experimental/__p1673_bits/conjugate_if_needed.hpp
```diff
@@ -0,0 +1,85 @@
...
+template<> struct is_complex<std::complex<const long double>> : std::true_type{};
+
+template<class T> inline constexpr bool is_complex_v = is_complex<T>::value;
+
+auto conj_if_needed = [](const auto value)
+{
+  using std::conj;
+  // we need to use remove_cv because in some cases, e.g. gcc10,
+  // decltype(value) carries a const qualification
+  if constexpr (is_complex_v<std::remove_cv_t<decltype(value)>>) {
```
mhoemmen, line 72: Please see note above -- your use of `std::remove_cv_t` means that you really only need to make `is_complex` inherit from `true_type` for `std::complex<float>`, `std::complex<double>`, and `std::complex<long double>`.  There is no `std::complex<const float>` etc.

mhoemmen (changes requested):
Only one change required -- please see note.  Thanks! : - )

include/experimental/__p1673_bits/conjugate_if_needed.hpp
mhoemmen: Short answer: Only make `is_complex<T>` inherit from `true_type` for `T` = `std::complex<float>`, `std::complex<double>`, and `std::complex<long double>`.

fnrizzi: @mhoemmen 
`decay` is going to remove the `const` from something like `const T`, decaying it to just `T`. 
It seems to me that you can instantiate complex with const-qualified value_type, and if we don't provide the specializations for `is_complex` the way I did, it won't work, see this: https://godbolt.org/z/YzTG4ovYT

mhoemmen: @fnrizzi *[complex.numbers.general]* 2 says: "The effect of instantiating the template complex for any type other than `float`, `double`, or `long double` is unspecified."  We shouldn't support undefined behavior that happens to work in some implementations.

fnrizzi: thanks @mhoemmen for pointing this out! Sorry for dragging the discussion... I did not know about this detail ! 
Now I know.

fnrizzi:
@mhoemmen fixed as per your suggestion!

mhoemmen (approved):

mhoemmen:
Thanks @fnrizzi ! : - D

PR #132, opened 2021-12-07
A few small fixes for compilation and installation

lums658:
This was to get compilation and installation working with g++-11 on MacOS.  Added a few "typename" here and there in a couple of files.  Also changed "cmake" to lib/cmake/stdBLAS for installing the stdBLAS cmake file(s).

I note that compilation is only possible with -std=c++20 now -- many uses of concepts and requires clauses throughout.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,7 +122,7 @@ install(TARGETS linalg EXPORT linalgTargets
 install(EXPORT linalgTargets
     FILE linalgTargets.cmake
     NAMESPACE std::
-    DESTINATION cmake
+    DESTINATION lib/cmake/stdBLAS
 )
 
 export(TARGETS linalg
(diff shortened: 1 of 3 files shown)


include/experimental/__p1673_bits/layout_blas_general.hpp
```diff
@@ -47,6 +47,10 @@
 #include "maybe_static_size.hpp"
 #include "layout_tags.hpp"
 
+#include <experimental/__p0009_bits/macros.hpp>
```
mhoemmen, line 50: The P1673 implementation shouldn't be depending on P0009 implementation details.  Would you consider instead just including the main `mdspan` header, <experimental/mdspan>?  Thanks!

mhoemmen (changes requested):
Thanks so much for the fixes! : - ) There's just one small thing; please see comment.

CMakeLists.txt
```diff
@@ -122,7 +122,7 @@ install(TARGETS linalg EXPORT linalgTargets
 install(EXPORT linalgTargets
     FILE linalgTargets.cmake
     NAMESPACE std::
-    DESTINATION cmake
+    DESTINATION lib/cmake/stdBLAS
```
dalg24, line 125: Should be `lib/cmake/<ProjectName>` so `stdBLAS -> linalg`?

mhoemmen: Ah, yes, that's right, thanks for the catch!

mhoemmen:
@lums658 I'm a bit occupied today, but if you're too busy to push the requested changes, would you mind if we did it?  Thanks!

PR #133, opened 2021-12-08
Two bug fixes in blas3_triangular_matrix_matrix_solve

youyu3:
* Drops const qualifier on temporal variables
* Corrects variable to be incremented to fix hang


mhoemmen (approved):
Thanks for the fixes! : - )

PR #135, opened 2021-12-15
Fix conjugated_scalar & accessor_conjugate

mhoemmen:
Fix conjugated_scalar & accessor_conjugate, so that they pass along the nested accessor.

This should fix #131.

PR #137, opened 2021-12-15
See if we can remove comparison operators from conjugated_scalar

mhoemmen:
This is otherwise the same as PR #135.


mhoemmen:
Turns out the comparison operators are useful for tests.  I'm leaving them for now.

Issue #134, opened 2021-12-08
P1673: Overloading of symmetric_matrix_rank_k_update and similar functions is ambiguous


mhoemmen:
@youyu3 observed that functions like `symmetric_matrix_rank_k_update` take all possible combinations of (`ExecutionPolicy&&` or not, `alpha` or not).  Without further constraints on `ExecutionPolicy`, this makes calls to the function ambiguous. 

This is an issue in P1673 itself; we'll need some way to resolve this.  The most straightforward way would be to require `alpha` unconditionally (even if it's always `1.0`), and only overload on `ExecutionPolicy&&`.

mhoemmen:
The issue with the proposal should be fixed in R6 (PR https://github.com/ORNL/cpp-proposals-pub/pull/186 ).  I'll file a separate issue for the implementation.

mhoemmen:
R6 was submitted with the overloads removed.  However, per discussion with @crtrott , I'm leaving this issue open.  We may want to bring back the overloads, if we can figure out how to constrain them (e.g., by saying that `alpha` has to work in a suitable arithmetic expression).

PR #139, opened 2021-12-15
rank_k_update: remove ambiguity, #134, #136

### LINKED ISSUE #136: Fix {symmetric,hermitian}_matrix_rank_k_update ambiguity
Remove all overloads of `symmetric_matrix_rank_k_update` and
`hermitian_matrix_rank_k_update` that do not take an `alpha` parameter.
This prevents ambiguity between overloads
that take `ExecutionPolicy&&` but not `alpha`,
and overloads that take `alpha` but not `ExecutionPolicy&&`.

This is a follow-on issue for https://github.com/kokkos/stdBLAS/issues/134 .  That issue relates to the proposal; this one relates to the implementation.

fnrizzi:
@mhoemmen this should fix #136


mhoemmen:
Thanks for the fix!  @crtrott asked us to investigate whether it's possible to constrain `alpha` so that we can keep the overloads.  Would you consider trying out this investigation?  I _think_ we would just have to add a `requires` clause that includes `alpha` in the algorithm's "mathematical expression" (that's a phrase I define in P1673, meaning "the expression that needs to compile and make semantic sense, for that algorithm").

fnrizzi:
@mhoemmen yes I definitely I am up to look at it if @crtrott approves me looking at it since he is in charge of my work :)

fnrizzi:
as soon as I move to blas3, i will take this on

PR #140, opened 2021-12-16
conjugated: Fixes operators and updates test

youyu3:


crtrott (approved):
Approving for now. Its clearly better than what we had. However we need to have a longer discussion about what scaled_scalar etc. are supposed to support.

include/experimental/__p1673_bits/conjugated.hpp
```diff
@@ -90,7 +86,12 @@ class conjugated_scalar {
 
 template<class T1, class Reference, class Element>
 auto operator* (const T1 x, const conjugated_scalar<Reference, Element> y) {
-  return x * Reference(y);
+  return x * Element(y);
```
mhoemmen, line 89: This should be fine -- if Reference is an expression type, then `Element(y)` would just force evaluation of the expression.

PR #151, opened 2022-03-30
[ready] tests: reorganize tests subdirs

fnrizzi:
@mhoemmen 
Here I just reorganized tests so that we can add the kokkos-based more cleanly, and added some things for the kokkos tests


mhoemmen:
@fnrizzi Do the tests still run without Kokkos as the back end?  I'm not quite clear why there needs to be a separate directory for that case.

mhoemmen (commented):
1. There's still some code in kokkos-based that looks not related to Kokkos-specific testing.  Is the plan to move that code later, or is it actually related to Kokkos-specific testing?
2. Why are the algorithm-specific tests in the "native" directory?  Wouldn't we need algorithm-specific tests for Kokkos-specific testing as well?

fnrizzi:
> 1. There's still some code in kokkos-based that looks not related to Kokkos-specific testing.  Is the plan to move that code later, or is it actually related to Kokkos-specific testing?
> 
All code inside Kokkos-based is needed for Kokkos testing and was written from scratch basically. 
All code inside native is simply a porting of the non-Kokkos code that was previously in the single test directory. 


> 2. Why are the algorithm-specific tests in the "native" directory?  Wouldn't we need algorithm-specific tests for Kokkos-specific testing as well?
All code inside native only tests the reference implementation. 
I have all algorithm-specific tests inside Kokkos-based too, I just need to move them but will do it one algo at a time to make it more readable. 

Does this address your concerns?

mhoemmen:
> I don't see why this is a problem? I feel like doing this makes things better organized.

I'm just trying to understand the changes : - )

mhoemmen (approved):

fnrizzi:
thanks @mhoemmen !

PR #160, opened 2022-04-08
[ready] `swap`: kokkos impl

fnrizzi:

### DIFF
--- a/tests/kokkos-based/swap_elements_rank1_kokkos.cpp
+++ b/tests/kokkos-based/swap_elements_rank1_kokkos.cpp
@@ -64,7 +64,7 @@ TEST_F(blas1_signed_complex_double_fixture, kokkos_swap)
 {
   using kc_t   = Kokkos::complex<double>;
   using stdc_t = value_type;
-  if (alignof(value_type) == alignof(kc_t)){
+  if constexpr(alignof(value_type) == alignof(kc_t)){
     kokkos_blas_swap_test_rank1_impl(x, y);
   }
 }
--- a/tests/kokkos-based/swap_elements_rank1_kokkos.cpp
+++ b/tests/kokkos-based/swap_elements_rank1_kokkos.cpp
@@ -1,0 +1,70 @@
+
+#include "gtest_fixtures.hpp"
+#include "helpers.hpp"
+
+namespace{
+
+template<class x_t, class y_t>
+void kokkos_blas_swap_test_rank1_impl(x_t x, y_t y)
+{
+  namespace stdla = std::experimental::linalg;
+
+  using value_type = typename x_t::value_type;
+  const std::size_t extent = x.extent(0);
+
+  // verify that x, y are different before running kernel
+  for (std::size_t i=0; i<extent; ++i){
+    EXPECT_TRUE(x(i)!=y(i));
+  }
+
+  // copy x and y before kernel
+  auto x_preKernel = kokkostesting::create_stdvector_and_copy(x);
+  auto y_preKernel = kokkostesting::create_stdvector_and_copy(y);
+
+  stdla::swap_elements(KokkosKernelsSTD::kokkos_exec<>(), x, y);
+
+  // after kernel, x should be unchanged, y should be equal to x
+  if constexpr(std::is_same_v<value_type, float>){
+    for (std::size_t i=0; i<extent; ++i){
+      EXPECT_FLOAT_EQ(x(i), y_preKernel[i]);
+      EXPECT_FLOAT_EQ(y(i), x_preKernel[i]);
+    }
+  }
+
+  if constexpr(std::is_same_v<value_type, double>){
+    for (std::size_t i=0; i<extent; ++i){
+      EXPECT_DOUBLE_EQ(x(i), y_preKernel[i]);
+      EXPECT_DOUBLE_EQ(y(i), x_preKernel[i]);
+    }
+  }
+
+  if constexpr(std::is_same_v<value_type, std::complex<double>>){
+    for (std::size_t i=0; i<extent; ++i){
+      EXPECT_DOUBLE_EQ(x(i).real(), y_preKernel[i].real());
+      EXPECT_DOUBLE_EQ(x(i).imag(), y_preKernel[i].imag());
+
+      EXPECT_DOUBLE_EQ(y(i).real(), x_preKernel[i].real());
+      EXPECT_DOUBLE_EQ(y(i).imag(), x_preKernel[i].imag());
+    }
+  }
+}
+}//end anonym namespace
+
+TEST_F(blas1_signed_float_fixture, kokkos_swap)
+{
+  kokkos_blas_swap_test_rank1_impl(x, y);
+}
+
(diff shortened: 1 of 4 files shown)


tests/kokkos-based/swap_elements_rank1_kokkos.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+  // after kernel, x should be unchanged, y should be equal to x
+  if constexpr(std::is_same_v<value_type, float>){
+    for (std::size_t i=0; i<extent; ++i){
+      EXPECT_FLOAT_EQ(x(i), y_preKernel[i]);
+      EXPECT_FLOAT_EQ(y(i), x_preKernel[i]);
+    }
+  }
+
+  if constexpr(std::is_same_v<value_type, double>){
```
mhoemmen, line 34: It's correct to use `value_type` here.  PR #162 (probably incorrectly) uses `element_type` instead.  (Please use `value_type` generally when you want to test the type with any `const` or `volatile` qualifications removed.)

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_swap_elements_kk.hpp
```diff
@@ -0,0 +1,94 @@
...
+#define LINALG_TPLIMPLEMENTATIONS_INCLUDE_EXPERIMENTAL___P1673_BITS_KOKKOSKERNELS_SWAP_HPP_
+
+#include <utility>
+
+namespace KokkosKernelsSTD {
+
+// this is here until we can use kokkos 3.6 which has swap avail
+template <class T>
+KOKKOS_INLINE_FUNCTION void _my_tmp_swap(T& a, T& b) noexcept {
+  static_assert(
```
mhoemmen, line 12: Per P1673, we need Constraints, not Mandates.  That is, we would want to use concepts or `enable_if_t` to constrain the template parameters; we wouldn't want to use `static_assert`.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_swap_elements_kk.hpp
```diff
@@ -0,0 +1,94 @@
...
+		     std::experimental::default_accessor<ElementType_x>
+		   > x,
+		   std::experimental::mdspan<
+		     ElementType_y,
+		     std::experimental::extents<ext_y ...>,
+		     Layout_y,
+		     std::experimental::default_accessor<ElementType_y>
+		   > y)
+{
+  // constraints
```
mhoemmen, line 48: "Constraints" are implemented through SFINAE, concepts, or `requires` clauses.  "Mandates" are implemented through `static_assert` or something like it.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_swap_elements_kk.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+  if constexpr(x.rank()==2){
+    if ( x.extent(1) != y.extent(1) ){
+      throw std::runtime_error("KokkosBlas: swap_elements: x.extent(1) != y.extent(1) ");
+    }
+  }
+
+  // this print is detected in the tests
+#if defined KOKKOS_STDBLAS_ENABLE_TESTS
+  std::cout << "swap_elements: kokkos impl\n";
```
mhoemmen, line 65: Would you consider encapsulating debug logging, rather than forcing include of `<iostream>`?  For example:
```c++
extern void logKokkos(const std::string_view functionName);
```
and then
```c++
logKokkos("swap_elements");
```

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_swap_elements_kk.hpp
```diff
@@ -0,0 +1,94 @@
...
+#endif
+
+  auto x_view = Impl::mdspan_to_view(x);
+  auto y_view = Impl::mdspan_to_view(y);
+
+  auto ex = ExeSpace();
+  if constexpr(x.rank()==1){
+    Kokkos::parallel_for(Kokkos::RangePolicy(ex, 0, x_view.extent(0)),
+			 KOKKOS_LAMBDA (const std::size_t & i){
```
mhoemmen, line 74: Please pass types with size <= `sizeof(void*)` by value, not by reference.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_swap_elements_kk.hpp
```diff
@@ -0,0 +1,94 @@
...
+			   _my_tmp_swap(x_view(i), y_view(i));
+			 });
+  }
+
+  else{
+    Kokkos::parallel_for(Kokkos::RangePolicy(ex, 0, x_view.extent(0)),
+			 KOKKOS_LAMBDA (const std::size_t & i){
```
mhoemmen, line 81: Please pass types with size <= `sizeof(void*)` by value, not by reference.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/mdspan_to_view_mapper_kk.hpp
```diff
@@ -22,19 +30,72 @@ struct LayoutMapper<std::experimental::layout_right> {
 };
 
 template<class MDSpan>
-std::enable_if_t<MDSpan::rank()==1, Kokkos::View<typename MDSpan::element_type*>>
+std::enable_if_t<
+  MDSpan::rank()==1
+  && !is_complex_v<typename MDSpan::element_type>,
```
mhoemmen, line 35: Please see my comments on PR #162: this should be `value_type` here, so that it works for cv-qualified types.

mhoemmen (changes requested):

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_swap_elements_kk.hpp
fnrizzi: yes totally right! Thanks. 
I need to get in the frame of mind of c++20 :) i am still mostly stuck at c++11/14 as you well know from sandia lagging

mhoemmen: Thanks @fnrizzi ! : - )  Constraints vs. Mandates isn't really a C++ >= 20 thing, but I totally understand needing to shift one's way of thinking.  I was lucky to be working for all but the last few months in a mostly C++20 code base.

fnrizzi:
@mhoemmen this is ready IMO

mhoemmen (approved):
Looks great!  Thanks for the contribution and for your responses to feedback!

It looks like there's a merge conflict, alas.

fnrizzi:
@mhoemmen conflict fixed via #172

PR #161, opened 2022-04-08
[ready] `scale`: kokkos impl

fnrizzi:

### DIFF
--- a/tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_scale_kk.hpp
+++ b/tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_scale_kk.hpp
@@ -3,17 +3,35 @@
 #define LINALG_TPLIMPLEMENTATIONS_INCLUDE_EXPERIMENTAL___P1673_BITS_KOKKOSKERNELS_SCALE_HPP_
 
 namespace KokkosKernelsSTD {
-template<class ExecSpace,
+
+//
+// for now, specialize for default_accessor
+// https://github.com/kokkos/stdBLAS/issues/122
+//
+template<class ExeSpace,
          class Scalar,
          class ElementType,
          std::experimental::extents<>::size_type ... ext,
-         class Layout,
-         class Accessor>
-void scale(kokkos_exec<ExecSpace>, const Scalar alpha,
-           std::experimental::mdspan<ElementType, std::experimental::extents<ext ...>, Layout, Accessor> x)
+         class Layout>
+void scale(kokkos_exec<ExeSpace> /*kexe*/,
+       const Scalar alpha,
+           std::experimental::mdspan<
+         ElementType,
+         std::experimental::extents<ext ...>,
+         Layout,
+         std::experimental::default_accessor<ElementType>
+       > obj)
 {
-    KokkosBlas::scal(Impl::mdspan_to_view(x),alpha,Impl::mdspan_to_view(x));
-}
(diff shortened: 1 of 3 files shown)


tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_scale_kk.hpp
```diff
@@ -3,17 +3,35 @@
...
+         std::experimental::default_accessor<ElementType>
+       > obj)
 {
-    KokkosBlas::scal(Impl::mdspan_to_view(x),alpha,Impl::mdspan_to_view(x));
-}
+  // constraints
+  static_assert(obj.rank() <= 2);
+
+#if defined KOKKOS_STDBLAS_ENABLE_TESTS
+  std::cout << "scale: kokkos impl\n";
```
mhoemmen, line 29: Please see my comment on e.g., PR #160 about encapsulating logging.  This is not required but would be a good idea (I think : - ) ).  Thanks!

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/mdspan_to_view_mapper_kk.hpp
```diff
@@ -22,19 +30,72 @@ struct LayoutMapper<std::experimental::layout_right> {
 };
 
 template<class MDSpan>
-std::enable_if_t<MDSpan::rank()==1, Kokkos::View<typename MDSpan::element_type*>>
+std::enable_if_t<
+  MDSpan::rank()==1
+  && !is_complex_v<typename MDSpan::element_type>,
```
mhoemmen, line 35: Please see my comments on PR #162.  This won't work for e.g., `mdspan<const std::complex<T>, ...>`.

mhoemmen (changes requested):

fnrizzi:
@mhoemmen this is ready IMO

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_scale_kk.hpp
```diff
@@ -2,18 +2,34 @@
...
+template<class ExeSpace,
          class Scalar,
          class ElementType,
          std::experimental::extents<>::size_type ... ext,
-         class Layout,
-         class Accessor>
-void scale(kokkos_exec<ExecSpace>, const Scalar alpha,
-           std::experimental::mdspan<ElementType, std::experimental::extents<ext ...>, Layout, Accessor> x)
+         class Layout>
+requires (sizeof...(ext) <= 2)
```
mhoemmen, line 18: This is a perfectly fine way to express the constraint.  You could also add a `requires` clause after the function's parameters, and do `requires(obj.rank() <= 2)`.

mhoemmen (approved):
Looks great!  Thanks for the contribution and for your responses to feedback!

It looks like there's a merge conflict, alas.

fnrizzi:
@mhoemmen conflict fixed via #172

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_scale_kk.hpp
mhoemmen: Just to clarify: Either way is fine with me : - )

mhoemmen (approved):

PR #162, opened 2022-04-08
mdspan_to_view_mapper: support complex type

fnrizzi:
this follows the slack discussion with Mark and Christian

### DIFF
--- a/tests/kokkos-based/mdspan_to_view.cpp
+++ b/tests/kokkos-based/mdspan_to_view.cpp
@@ -2,6 +2,12 @@
 #include "gtest_fixtures.hpp"
 #include "helpers.hpp"
 
+template<class MDSpanType, class KViewType>
+void expect_shallow_copy(MDSpanType mdsp, KViewType kv)
+{
+  EXPECT_EQ( (void *) mdsp.data(), (void *) kv.data() );
+}
+
 template<class MDSpanValueType, class KViewValueType = MDSpanValueType>
 void mdspan_to_view_test_impl()
 {
@@ -20,6 +26,7 @@ void mdspan_to_view_test_impl()
     static_assert(kv_type::rank == 1);
     static_assert(std::is_same_v<typename kv_type::value_type, KViewValueType>);
     EXPECT_TRUE(kv.extent(0) == 5);
+    expect_shallow_copy(mdsp, kv);
   }
 
   // rank1, const
@@ -33,6 +40,7 @@ void mdspan_to_view_test_impl()
     static_assert(kv_type::rank == 1);
     static_assert(std::is_same_v<typename kv_type::value_type, const KViewValueType>);
     EXPECT_TRUE(kv.extent(0) == 5);
+    expect_shallow_copy(mdsp, kv);
   }
 
   // rank2, non-const
@@ -47,6 +55,7 @@ void mdspan_to_view_test_impl()
     static_assert(std::is_same_v<typename kv_type::value_type, KViewValueType>);
     EXPECT_TRUE(kv.extent(0) == 3);
     EXPECT_TRUE(kv.extent(1) == 4);
+    expect_shallow_copy(mdsp, kv);
   }
 
   // rank2, const
@@ -61,6 +70,7 @@ void mdspan_to_view_test_impl()
     static_assert(std::is_same_v<typename kv_type::value_type, const KViewValueType>);
     EXPECT_TRUE(kv.extent(0) == 3);
     EXPECT_TRUE(kv.extent(1) == 4);
+    expect_shallow_copy(mdsp, kv);
   }
 }
 
--- a/tests/kokkos-based/mdspan_to_view.cpp
+++ b/tests/kokkos-based/mdspan_to_view.cpp
@@ -1,0 +1,81 @@
+
+#include "gtest_fixtures.hpp"
+#include "helpers.hpp"
+
+template<class MDSpanValueType, class KViewValueType = MDSpanValueType>
+void mdspan_to_view_test_impl()
+{
+  using std::experimental::mdspan;
+  using std::experimental::extents;
+  using std::experimental::dynamic_extent;
+
+  // rank1, non-const
+  {
+    std::vector<MDSpanValueType> a(5);
+    using mdspan_t = mdspan<MDSpanValueType, extents<dynamic_extent>>;
+    mdspan_t mdsp(a.data(), a.size());
+
+    auto kv = KokkosKernelsSTD::Impl::mdspan_to_view(mdsp);
+    using kv_type = decltype(kv);
+    static_assert(kv_type::rank == 1);
+    static_assert(std::is_same_v<typename kv_type::value_type, KViewValueType>);
+    EXPECT_TRUE(kv.extent(0) == 5);
+  }
+
+  // rank1, const
+  {
+    std::vector<MDSpanValueType> a(5);
+    using mdspan_t = mdspan<const MDSpanValueType, extents<dynamic_extent>>;
+    mdspan_t mdsp(a.data(), a.size());
+
+    auto kv = KokkosKernelsSTD::Impl::mdspan_to_view(mdsp);
+    using kv_type = decltype(kv);
+    static_assert(kv_type::rank == 1);
+    static_assert(std::is_same_v<typename kv_type::value_type, const KViewValueType>);
+    EXPECT_TRUE(kv.extent(0) == 5);
+  }
+
+  // rank2, non-const
+  {
+    std::vector<MDSpanValueType> a(12);
+    using mdspan_t = mdspan<MDSpanValueType, extents<dynamic_extent, dynamic_extent>>;
+    mdspan_t mdsp(a.data(), 3, 4);
+
+    auto kv = KokkosKernelsSTD::Impl::mdspan_to_view(mdsp);
+    using kv_type = decltype(kv);
+    static_assert(kv_type::rank == 2);
+    static_assert(std::is_same_v<typename kv_type::value_type, KViewValueType>);
+    EXPECT_TRUE(kv.extent(0) == 3);
+    EXPECT_TRUE(kv.extent(1) == 4);
+  }
+
+  // rank2, const
+  {
+    std::vector<MDSpanValueType> a(12);
+    using mdspan_t = mdspan<const MDSpanValueType, extents<dynamic_extent, dynamic_extent>>;
+    mdspan_t mdsp(a.data(), 3, 4);
+
+    auto kv = KokkosKernelsSTD::Impl::mdspan_to_view(mdsp);
+    using kv_type = decltype(kv);
+    static_assert(kv_type::rank == 2);
+    static_assert(std::is_same_v<typename kv_type::value_type, const KViewValueType>);
+    EXPECT_TRUE(kv.extent(0) == 3);
+    EXPECT_TRUE(kv.extent(1) == 4);
+  }
+}
+
+TEST(mdspan_to_view, for_float){
+  mdspan_to_view_test_impl<float>();
+}
+
+TEST(mdspan_to_view, for_double){
+  mdspan_to_view_test_impl<double>();
+}
+
+TEST(mdspan_to_view, for_complex_double){
+  using value_type = std::complex<double>;
+  using kc_t       = Kokkos::complex<double>;
+  if constexpr (alignof(value_type) == alignof(kc_t)){
+    mdspan_to_view_test_impl<value_type, kc_t>();
+  }
+}
--- a/tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/mdspan_to_view_mapper_kk.hpp
+++ b/tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/mdspan_to_view_mapper_kk.hpp
@@ -5,12 +5,12 @@
 namespace KokkosKernelsSTD {
 namespace Impl {
 
-template<class T> struct is_complex : std::false_type{};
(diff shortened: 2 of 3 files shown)


tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/mdspan_to_view_mapper_kk.hpp
```diff
@@ -22,19 +30,72 @@ struct LayoutMapper<std::experimental::layout_right> {
 };
 
 template<class MDSpan>
-std::enable_if_t<MDSpan::rank()==1, Kokkos::View<typename MDSpan::element_type*>>
+std::enable_if_t<
+  MDSpan::rank()==1
+  && !is_complex_v<typename MDSpan::element_type>,
```
mhoemmen, line 35: This will fail for `mdspan<const std::complex<T>, ...>`.  If that's not the intent, then a potential fix would be to use `MDSpan::value_type` (which removes any `const` and `volatile` qualifications) instead of `MDSpan::element_type`.  You would need to use `value_type` instead of `element_type` everywhere you use `is_complex_v`.

Another option would be to define `is_complex_v` to use `remove_cv_t` on its argument.  I personally wouldn't mind that, but my view is that using `MDSpan::value_type` better expresses the desired intent, which is to say something about the value type (with any `const` or `volatile` qualifiers removed).

In my view, an even better approach would be not to specialize `mdspan_to_view` on whether the value type is complex.  Instead, you could write a function just to convert the input's raw pointer to a raw pointer that Kokkos understands, and then use that function everywhere.  Godbolt link with an example: https://godbolt.org/z/Pb39fq1so

```c++
template<class ValueType>
ValueType* toKokkosPointer(ValueType* p) { return p; }

template<class RealType>
auto toKokkosPointer(std::complex<RealType>* p) {
    assertAlignment(p); // whatever this should be
    return reinterpret_cast<Kokkos::complex<RealType>*>(p);
}
```

There are two reasons to do it this way.

1. It orthogonalizes different reasons to specialize into different layers.  One layer handles raw pointer conversion, and the other layer handles rank 1, rank 2, rank 3, etc.
2. It opens up an extension point for other kinds of pointer conversions that you might need in the future.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/mdspan_to_view_mapper_kk.hpp
```diff
@@ -22,19 +30,72 @@ struct LayoutMapper<std::experimental::layout_right> {
...
+  MDSpan::rank()==1
+  && is_complex_v<typename MDSpan::element_type>,
+  Kokkos::View< Kokkos::complex<typename MDSpan::element_type::value_type>* >
+  >
+mdspan_to_view(const MDSpan& a)
+{
+  using mdspan_v_type = typename MDSpan::element_type::value_type;
+  using kc_t = Kokkos::complex<mdspan_v_type>;
+  if (alignof(mdspan_v_type) != alignof(kc_t)){
+    throw std::runtime_error("KokkosBlas: mdspan_to_view: mismatching alignment of std::complex<> and Kokkos::complex<> which prevents interoperability. By default, Kokkos::complex<T> is aligned as 2*sizeof(T) but std::complex<T> is aligned as sizeof(T). To match the alignement and allow correct conversion, you can rebuild Kokkos undefining KOKKOS_ENABLE_COMPLEX_ALIGN.");
```
mhoemmen, line 53: Excellent error message!  Just a couple suggestions:
```suggestion
    throw std::runtime_error("KokkosBlas: mdspan_to_view: mismatching alignment of std::complex<> and Kokkos::complex<> which prevents interoperability. By default, Kokkos::complex<T> is aligned as 2*sizeof(T) but std::complex<T> is aligned as sizeof(T). To match the alignment and allow correct conversion, you must rebuild Kokkos with KOKKOS_ENABLE_COMPLEX_ALIGN undefined.");
```

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/mdspan_to_view_mapper_kk.hpp
```diff
@@ -22,19 +30,72 @@ struct LayoutMapper<std::experimental::layout_right> {
...
+    Kokkos::complex< typename MDSpan::element_type::value_type> **,
+    typename LayoutMapper<typename MDSpan::layout_type>::type
+    >
+  >
+mdspan_to_view(const MDSpan& a)
+{
+  using mdspan_v_type = typename MDSpan::element_type::value_type;
+  using kc_t = Kokkos::complex<mdspan_v_type>;
+  if (alignof(mdspan_v_type) != alignof(kc_t)){
+    throw std::runtime_error("KokkosBlas: mdspan_to_view: mismatching alignment of std::complex<> and Kokkos::complex<> which prevents interoperability. By default, Kokkos::complex<T> is aligned as 2*sizeof(T) but std::complex<T> is aligned as sizeof(T). To match the alignement and allow correct conversion, you can rebuild Kokkos undefining KOKKOS_ENABLE_COMPLEX_ALIGN.");
```
mhoemmen, line 92: Please see above note.  Also, would you consider consolidating the check and error message into a function like `assertAlignment`?

mhoemmen (changes requested):
There's a bug, as far as I can tell: The code won't correctly make a `Kokkos::View<const Kokkos::complex<T>*, ...>` out of an `mdspan<const std::complex<T>, ...>`.  Please see suggestions below.  Thanks!

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/mdspan_to_view_mapper_kk.hpp
fnrizzi: nice suggestion, I followed it in this new revised impl !

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/mdspan_to_view_mapper_kk.hpp
fnrizzi: done!

fnrizzi:
@mhoemmen what do you think of my revised implementation? I also added a test

tests/kokkos-based/mdspan_to_view.cpp
```diff
@@ -0,0 +1,81 @@
...
+  {
+    std::vector<MDSpanValueType> a(5);
+    using mdspan_t = mdspan<MDSpanValueType, extents<dynamic_extent>>;
+    mdspan_t mdsp(a.data(), a.size());
+
+    auto kv = KokkosKernelsSTD::Impl::mdspan_to_view(mdsp);
+    using kv_type = decltype(kv);
+    static_assert(kv_type::rank == 1);
+    static_assert(std::is_same_v<typename kv_type::value_type, KViewValueType>);
+    EXPECT_TRUE(kv.extent(0) == 5);
```
mhoemmen, line 22: (OPTIONAL) Would you consider also adding the following tests to each case?

1. `EXPECT_EQ(mdsp.data(), kv.data())` (check that it's really a shallow copy)
2. Whether the `const`-ness of the Kokkos::View's element type is as expected (I think you do that below)

Thanks!

tests/kokkos-based/mdspan_to_view.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+TEST(mdspan_to_view, for_double){
+  mdspan_to_view_test_impl<double>();
+}
+
+TEST(mdspan_to_view, for_complex_double){
+  using value_type = std::complex<double>;
+  using kc_t       = Kokkos::complex<double>;
+  if constexpr (alignof(value_type) == alignof(kc_t)){
```
mhoemmen, line 78: Even if this isn't true, shouldn't users still be able to do `mdspan`-to-`Kokkos::View` conversions if they align their data manually?  In that case, wouldn't it be good still to have the test?

Changing it back to a run-time check is optional for approval.  I think it's OK to leave this for now and deal with it later only if needed.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/mdspan_to_view_mapper_kk.hpp
```diff
@@ -21,20 +21,56 @@ struct LayoutMapper<std::experimental::layout_right> {
   using type = Kokkos::LayoutRight;
 };
 
+template<class T>
+constexpr void static_assert_complex_alignment(const T* /* p */)
+{
+  using kc_t = Kokkos::complex<typename T::value_type>;
+  static_assert(alignof(T) == alignof(kc_t),
```
mhoemmen, line 28: Please see comment above.  Users should still be able to create a `Kokkos::View` from complex data if they align their data manually.  Doing a run-time test would, of course, be a bit harder on device, but Kokkos users generally aren't letting `std::complex` data exist in device code.

mhoemmen (approved):
Thank you for addressing the concerns expressed in my last review! : - )  I'm happy to merge this if you're done, though I have a couple suggestions (please see in-line comments).  Thanks so much for your contributions!

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/mdspan_to_view_mapper_kk.hpp
```diff
@@ -22,19 +30,72 @@ struct LayoutMapper<std::experimental::layout_right> {
 };
 
 template<class MDSpan>
-std::enable_if_t<MDSpan::rank()==1, Kokkos::View<typename MDSpan::element_type*>>
+std::enable_if_t<
+  MDSpan::rank()==1
+  && !is_complex_v<typename MDSpan::element_type>,
+  Kokkos::View<typename MDSpan::element_type*>
+  >
 mdspan_to_view(const MDSpan& a) {
```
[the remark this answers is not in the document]
fnrizzi: done!

tests/kokkos-based/mdspan_to_view.cpp
fnrizzi: done!

tests/kokkos-based/mdspan_to_view.cpp
fnrizzi: I thought that for now we wanted to restrict this to make sure Kokkos is built with the right alignment for complex.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/mdspan_to_view_mapper_kk.hpp
```diff
@@ -21,20 +21,56 @@ struct LayoutMapper<std::experimental::layout_right> {
...
+
+//
+// mdspan_to_view
+//
 template<class MDSpan>
-std::enable_if_t<MDSpan::rank()==1, Kokkos::View<typename MDSpan::element_type*>>
-mdspan_to_view(const MDSpan& a) {
-    return Kokkos::View<typename MDSpan::element_type*>(a.data(),a.extent(0));
+requires (MDSpan::rank()==1)
+auto mdspan_to_view(const MDSpan& a) {
```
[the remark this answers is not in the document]
fnrizzi: Yes good point, I changed it to be explicitly a mdspan. I totally agree that if make this generic, the name should be changed. So since we keep the name for now, I changed it so that it actually reflects that.

mhoemmen (approved):

tests/kokkos-based/mdspan_to_view.cpp
mhoemmen: I guess that's OK, though this test will just be skipped if Kokkos was not built with the right alignment.  Were you thinking of having the test fail otherwise?

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/mdspan_to_view_mapper_kk.hpp
```diff
@@ -21,20 +21,78 @@ struct LayoutMapper<std::experimental::layout_right> {
...
+  auto kokkos_p = to_kokkos_pointer(a.data());
+  using view_type = Kokkos::View<
+    decltype(kokkos_p)*,
+    typename LayoutMapper<typename mdspan_type::layout_type>::type
+    >;
+  return view_type(kokkos_p, a.extent(0), a.extent(1));
+}
+
 } // namespace Impl
 } // namespac KokkosKernelsSTD
```
MikolajZuzek, line 97: Typo: `namespac` (caught on https://github.com/kokkos/stdBLAS/pull/170#discussion_r848535016)

mhoemmen:
I think this is fine to merge, though I'll wait until you're ready.  You might like to fix the (minor) typo too, but we can always come back and fix that later.

tests/kokkos-based/mdspan_to_view.cpp
fnrizzi: i intended for this to be skipped if kokkos is not configured accordingly. But I can change it later to fail if Kokkos team wants that

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/mdspan_to_view_mapper_kk.hpp
fnrizzi: fixed!

fnrizzi:
@mhoemmen typo fixed, i am ready ! if you are happy with this

mhoemmen:
Thanks for the contribution!

PR #163, opened 2022-04-08
[ready] `matrix_vector_product`: kokkos impl

fnrizzi:


tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_gemv_kk.hpp
```diff
@@ -0,0 +1,228 @@
+
```
mhoemmen, line 1: I'm not sure if I've mentioned it on your other PRs, but would you consider adding a copyright notice to the top of this new file?

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_gemv_kk.hpp
```diff
@@ -0,0 +1,228 @@
...
+			   > x,
+			   std::experimental::mdspan<
+			     ElementType_y,
+			     std::experimental::extents<ext_y>,
+			     Layout_y,
+			     std::experimental::default_accessor<ElementType_y>
+			   > y)
+{
+
+  // constraints
```
mhoemmen, line 55: "Constraints" are enforced through SFINAE, concepts, etc.  "Mandates" are enforced via `static_assert`.  (I'll take this for now, if the comment says something like "FIXME: enforce constraints as such, not as mandates.")

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_gemv_kk.hpp
```diff
@@ -0,0 +1,228 @@
...
+  if ( A.extent(1) != x.extent(0) ){
+    throw std::runtime_error("KokkosBlas: matrix_vector_product: A.extent(1) != x.extent(0) ");
+  }
+  if ( A.extent(0) != y.extent(0) ){
+    throw std::runtime_error("KokkosBlas: matrix_vector_product: A.extent(0) != y.extent(0) ");
+  }
+
+  // this print is detected in the tests
+#if defined KOKKOS_STDBLAS_ENABLE_TESTS
+  std::cout << "overwriting_matrix_vector_product: kokkos impl\n";
```
mhoemmen, line 70: I've made this comment on other PRs related to this one: Would you consider encapsulating logging, so that mdspan files aren't exposed to `#include <iostream>`?
```c++
extern void logKokkos(const std::string_view functionName);
```
with `logKokkos` defined in a .cpp file somewhere, would make it possible to log in different ways, without needing to change existing code.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_gemv_kk.hpp
```diff
@@ -0,0 +1,228 @@
...
+#endif
+
+  auto A_view = Impl::mdspan_to_view(A);
+  auto x_view = Impl::mdspan_to_view(x);
+  auto y_view = Impl::mdspan_to_view(y);
+
+  // make alpha and beta consistent with types in KokkosBals::gemv
```
mhoemmen, line 77:
```suggestion
  // make alpha and beta consistent with types in KokkosBlas::gemv
```

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_gemv_kk.hpp
```diff
@@ -0,0 +1,228 @@
...
+			   > y,
+			   std::experimental::mdspan<
+			     ElementType_z,
+			     std::experimental::extents<ext_z>,
+			     Layout_z,
+			     std::experimental::default_accessor<ElementType_z>
+			   > z)
+{
+
+  // constraints
```
mhoemmen, line 130: Please see above comment on Constraints.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_gemv_kk.hpp
```diff
@@ -0,0 +1,228 @@
...
+  auto A_view = Impl::mdspan_to_view(A);
+  auto x_view = Impl::mdspan_to_view(x);
+  auto y_view = Impl::mdspan_to_view(y);
+  auto z_view = Impl::mdspan_to_view(z);
+
+  // FRIZZI: we need to improve this or maybe we should
+  // just do two ops: fist y->z, and then gemv directly
+  auto ex = ExeSpace();
+  Kokkos::parallel_for(Kokkos::RangePolicy(ex, 0, z_view.extent(0)),
+		       KOKKOS_LAMBDA (const std::size_t & i)
```
mhoemmen, line 161: Would you consider passing this index by value instead of by const reference?  On computers with 64-bit pointers, there's no benefit to using a const reference here.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_gemv_kk.hpp
```diff
@@ -0,0 +1,228 @@
...
+		        ElementType_y,
+		        std::experimental::extents<ext_y>,
+		        Layout_y,
+		        Accessor_y
+		      > y)
+{
+  auto alpha = gemvimpl::get_scaling_factor(A.accessor())  *
+               gemvimpl::get_scaling_factor(x.accessor());
+  KokkosBlas::gemv("N", alpha, Impl::mdspan_to_view(A),
+		   Impl::mdspan_to_view(x), 0.0,
```
mhoemmen, line 222: Would you consider making sure that beta has the same type as `alpha` (or some other appropriate type), instead of just using `double` (i.e., `0.0`) here?

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/mdspan_to_view_mapper_kk.hpp
```diff
@@ -22,19 +30,72 @@ struct LayoutMapper<std::experimental::layout_right> {
 };
 
 template<class MDSpan>
-std::enable_if_t<MDSpan::rank()==1, Kokkos::View<typename MDSpan::element_type*>>
+std::enable_if_t<
+  MDSpan::rank()==1
+  && !is_complex_v<typename MDSpan::element_type>,
```
mhoemmen, line 35: Please see comments on e.g., PR #162.  This at least should be `value_type` instead of `element_type`, else it won't work for `mdspan<const std::complex<T>, ...>`.

mhoemmen (changes requested):

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_gemv_kk.hpp
fnrizzi: Thanks for pointing it out, I can work on changing it. 
Btw, off the top of my head I think there are a few places I saw inside the reference implementation where the same inconsistency is present, namely static asserts for constraints, but i can double check.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_gemv_kk.hpp
fnrizzi: nice one, thanks !

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_gemv_kk.hpp
fnrizzi: yes, sorry, this slipped! thanks for bringing this up

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_gemv_kk.hpp
mhoemmen: The reference implementation needs more work, for sure : - )  I'm just hoping that new code can Do The Right Thing, so we have less old code to fix in the future.  Thanks!

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_gemv_kk.hpp
```diff
@@ -0,0 +1,231 @@
...
+#ifndef LINALG_TPLIMPLEMENTATIONS_INCLUDE_EXPERIMENTAL___P1673_BITS_KOKKOSKERNELS_GEMV_HPP_
+#define LINALG_TPLIMPLEMENTATIONS_INCLUDE_EXPERIMENTAL___P1673_BITS_KOKKOSKERNELS_GEMV_HPP_
+
+#include "signal_kokkos_impl_called.hpp"
+
+namespace KokkosKernelsSTD {
+
+namespace gemv_impl{
+template<class Accessor>
+double get_scaling_factor(Accessor) { return 1.0; }
```
mhoemmen, line 11: This will force the scaling factor to have type `double`, even if no floating-point values are involved.  (What if all the matrices have element type `int64_t`, for example?)

Would you consider having `get_scaling_factor` take both an accessor type and a value type?  The value type could come from the calling context.  For example, in `y = A * x`, the scaling factor could have type `declval<typename decltype(A)::value_type>() * declval<typename decltype(x)::value_type>()`.

mhoemmen (changes requested):

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_gemv_kk.hpp
fnrizzi: @mhoemmen this code is here because it was added by Christian, so I did not touch it :) 
I can make the changes you suggest but just wanted to point out I have not touched this code for that reason!

fnrizzi: in fact, I also left the original specialization, see last overload in that header:
```
//
// the following specialization is here to avoid breaking
// the example code started by CTrott in this file:
//    stdBLAS/examples/kokkos-based/matrix_vector_product_kokkos.cpp
// but we need to change this later after we properly handle accessors
```

fnrizzi: @mhoemmen i made some changes

mhoemmen (approved):

crtrott (approved):

PR #164, opened 2022-04-08
[ready] `symmetric_matrix_vector_product`: kokkos impl

fnrizzi:


tests/kokkos-based/overwriting_symmetric_matrix_vector_product.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+  // copy operands before running the kernel
+  auto A_preKernel = kokkostesting::create_stdvector_and_copy_rowwise(A);
+  auto x_preKernel = kokkostesting::create_stdvector_and_copy(x);
+  auto y_preKernel = kokkostesting::create_stdvector_and_copy(y);
+
+  // compute y gold symv
+  std::vector<value_type> gold(y.extent(0));
+  using mdspan_t = mdspan<value_type, extents<dynamic_extent>>;
+  mdspan_t y_gold(gold.data(), y.extent(0));
```
mhoemmen, line 36: This code is fine and doesn't need to change.  However, given that we're assuming C++20 anyway, one could use CTAD here:
```c++
mdspan y_gold(gold.data(), y.extent(0));
```
I'm not sure how far we want to back-port the reference implementation of P1673, though.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_symv_kk.hpp
```diff
@@ -0,0 +1,253 @@
...
+  auto y_view = Impl::mdspan_to_view(y);
+
+  auto ex = ExeSpace();
+
+  if constexpr (std::is_same_v<Triangle, std::experimental::linalg::upper_triangle_t>)
+  {
+
+  // this print is detected in the tests
+#if defined KOKKOS_STDBLAS_ENABLE_TESTS
+  std::cout << "overwriting_symmetric_matrix_vector_product_upper: kokkos impl\n";
```
mhoemmen, line 73: Please see comments on my other PR about logging encapsulation; thanks! : - )

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/mdspan_to_view_mapper_kk.hpp
```diff
@@ -22,19 +30,72 @@ struct LayoutMapper<std::experimental::layout_right> {
 };
 
 template<class MDSpan>
-std::enable_if_t<MDSpan::rank()==1, Kokkos::View<typename MDSpan::element_type*>>
+std::enable_if_t<
+  MDSpan::rank()==1
+  && !is_complex_v<typename MDSpan::element_type>,
```
mhoemmen, line 35: Please see my comments on PR #162 (summary: this won't work for views of const; one way to fix that would be to use `value_type` here instead of `element_type` here).

mhoemmen (changes requested):

tests/kokkos-based/overwriting_symmetric_matrix_vector_product.cpp
fnrizzi: I can leave this for now and revise them at the end ok? since these tests will end up in Kokkos kernels, not sure we want to restrict ourselves to c++20

mhoemmen (approved):

tests/kokkos-based/overwriting_symmetric_matrix_vector_product.cpp
mhoemmen: @fnrizzi Right, that's a good point -- it would be helpful if the tests didn't require C++20.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_symv_kk.hpp
```diff
@@ -0,0 +1,251 @@
...
+  }
+  if ( A.extent(0) != y.extent(0) ){
+    throw std::runtime_error("KokkosBlas: matrix_vector_product: A.extent(0) != y.extent(0) ");
+  }
+
+  symv_impl::static_extent_match(A.static_extent(0), A.static_extent(1));
+  symv_impl::static_extent_match(A.static_extent(1), x.static_extent(0));
+  symv_impl::static_extent_match(A.static_extent(0), x.static_extent(0));
+
+  auto A_view = Impl::mdspan_to_view(A);
```
mhoemmen, line 73: @fnrizzi @MikolajZuzek 

This reminds me that the spec doesn't say anything about not letting `A` be a transposed layout.  I think that's OK as far as the spec is concerned, because the algorithm will just access the set of `(i,j)` corresponding to the `Triangle` argument.  However, `mdspan_to_view` will need to handle this case.

fnrizzi: i opened this: https://github.com/kokkos/stdBLAS/issues/175

fnrizzi: one thought: when Kokkos view will be based on mdspan, I assume both layout, accessors will carry over right?

mhoemmen: @fnrizzi yup!  I haven't looked at @crtrott 's efforts on making Kokkos::View use mdspan, but I can imagine a couple different options for the accessor.  It might not necessarily be `default_accessor`.

mhoemmen: It's OK to address this later with a fix for #175.

mhoemmen (approved):

PR #157, opened 2022-04-07
`matrix_rank_1_update`: kokkos impl

MikolajZuzek:
This is Kokkos implementation of `matrix_rank_1_update`, including conjugate overload for `matrix_rank_1_update_c`.
 #162  is a prerequisite for this, I can rebase once that is merged

PR #166, opened 2022-04-08
[ready] `hermitian_matrix_vector_product`: kokkos impl

fnrizzi:
- #162  is a prerequisite for this, I can rebase once that is merged. 
 - #164 is also a prerequisite for this because in the case we have a "real" value type, the `hemv` impl will just call `symv` as follows: 
 ```cpp
  if constexpr (hermitianmatvecimpl::is_complex_v<ElementType_A> == false){
    std::experimental::linalg::symmetric_matrix_vector_product(kexe, A, tr, x, y);
  }
```
 So `symv` has to be present for this to work correctly.

### CODE REVIEW

File: tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_hemv_kk.hpp [hunk 1]
```diff
@@ -0,0 +1,290 @@
...
+					std::experimental::default_accessor<ElementType_x>
+				      > x,
+				      std::experimental::mdspan<ElementType_y,
+					std::experimental::extents<ext_y>,
+					Layout_y,
+					std::experimental::default_accessor<ElementType_y>
+				      > y)
+{
+
+  // constraints
```
mhoemmen, line 57: Please see my comments elsewhere on the difference between Constraints and Mandates; thanks!

mhoemmen: I realize also that these `static_assert`s will never fail, because this overload only participates in overload resolution if they are all true.  Thus, I think I see what you're trying to do, which is double-check your constraints (rather than impose new Mandates conditions).  That's not bad, but it's unnecessary; the properties already hold by construction.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_hemv_kk.hpp
```diff
@@ -0,0 +1,290 @@
...
+      auto y_view = Impl::mdspan_to_view(y);
+
+      auto ex = ExeSpace();
+
+      if constexpr (std::is_same_v<Triangle, std::experimental::linalg::upper_triangle_t>)
+      {
+
+	// this print is detected in the tests
+#if defined KOKKOS_STDBLAS_ENABLE_TESTS
+	std::cout << "overwriting_hermitian_matrix_vector_product_upper: kokkos impl\n";
```
mhoemmen, line 96: Please see my comments elsewhere on encapsulating logging; thanks!

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_hemv_kk.hpp
```diff
@@ -0,0 +1,290 @@
...
+#endif
+
+	Kokkos::parallel_for(Kokkos::RangePolicy(ex, 0, A_view.extent(0)),
+			     KOKKOS_LAMBDA (const std::size_t & i)
```
mhoemmen, line 100: Please see my comments elsewhere on preferring to pass types of size less than or equal to `sizeof(void*)` by value, instead of by const reference; thanks!

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_hemv_kk.hpp
```diff
@@ -0,0 +1,290 @@
...
+    throw std::runtime_error("KokkosBlas: hermitian_matrix_vector_product: A.extent(0) != y.extent(0) ");
+  }
+
+  // for non-complex ELementType_A, A can be treated as symmetric
+  // http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p1673r6.html#hermitian-matrix-vector-product-linalgalgsblas2hemv
+  //
+  // If inout_matrix_t::element_type is complex<RA> for some RA, then ....
+  // Otherwise, the functions will assume that A[j,i] equals A[i,j].
+  //
+  if constexpr (hermitianmatvecimpl::is_complex_v<ElementType_A> == false){
```
mhoemmen, line 79: This will not have the desired results if, e.g., `ElementType_A` is `const std::complex<T>` for some `T`.  I've discussed this in my comments on PR #162, among other places.  One fix would be to use the `mdspan`'s `value_type` alias.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_hemv_kk.hpp
```diff
@@ -0,0 +1,290 @@
...
+			     {
+
+			       using lsum_type = decltype( A_view(0,0) * x_view(0) );
+			       lsum_type lsum = {};
+
+			       for (std::size_t j = i; j < A_view.extent(1); ++j) {
+				 lsum += A_view(i,j) * x_view(j);
+			       }
+			       for (std::size_t j = 0; j < i; ++j) {
+				 lsum += Kokkos::conj(A_view(j,i)) * x_view(j);
```
mhoemmen, line 110: Would you consider instead using a `conj_if_needed` function here, to make the function have a single representation for both complex and non-complex cases?

mhoemmen (changes requested):
Thanks for this PR!  The only required changes relate to PR #162 and the use of `value_type` rather than `element_type` for determining whether an `mdspan` is a view of complex elements.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_hemv_kk.hpp
MikolajZuzek: Would it make sense to follow your suggestion by replacing these `static_assert`s with something like:
```cpp
template<...>
  // P1673 constraints
  requires (requires () { decltype(A)::rank() == 2; })
        && (requires () { decltype(x)::rank() == 1; })
        && (requires () { decltype(y)::rank() == 1; })
void hermitian_matrix_vector_product(...)
```
?

It duplicates what `mdspan`/`extents` do in the function header now, but:
1. gives clear reference to and description of specification _Constraints_;
2. will alarm if someone unconsciously relaxes the constraints by changing the header (e.g. generalizing the types).

Alternatively, I thought of a comment like 
```cpp
// Note: P1673 constraints (rank 2 for A and rank 1 for x and y)
// are specified with mdspan<..., extents<*>, ...> in the function header
template<...>
void hermitian_matrix_vector_product(...)
```
but `requires` seems much better.

> PS. You touched on asserts duplicating the overload spec, so I'm asking here in context of https://github.com/kokkos/stdBLAS/pull/157#discussion_r846881993

mhoemmen: @MikolajZuzek Thanks for the suggestions! : - )  I would actually prefer nothing at all -- neither option -- based on the Don't Repeat Yourself (DRY) Principle.  The function's parameters already express the desired constraints.  If those ever need to change, then you would only have one place to change them.

If I had to pick one of the two options, it would be the `static_assert`s.  Adding redundant `requires` clauses would make debugging harder.  (It would actually shift debugging to run time, because you'll need the logging output to see that you're not getting the kokkos-kernels overload.)

Btw, the `requires` clauses can be spelled more concisely as (e.g.,) `requires(A.rank() == 2)`.  [Here](https://godbolt.org/z/aWj8e7Yvs) is an example.
```c++
#include <array>

template<class T>
T f(T t) requires(t.size() == 3)
{
    return t;
}

int main() {
    std::array<int, 3> x;
    std::array<int, 4> y;
    std::array<int, 5> z;

    auto x2 = f(x);
    //auto y2 = f(y); // no matching function
    //auto z2 = f(z); // no matching function

    return 0;
}
```

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_hemv_kk.hpp
fnrizzi: done

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_hemv_kk.hpp
fnrizzi: this is already inside the: 
```cpp
if constexpr (hemv_impl::is_std_complex_v<typename decltype(A)::value_type> == false){
}
else{
}
```
so we know A is complex

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_hemv_kk.hpp
```diff
@@ -0,0 +1,268 @@
...
+    throw std::runtime_error("KokkosBlas: hermitian_matrix_vector_product: A.extent(0) != A.extent(1) ");
+  }
+  if ( A.extent(1) != x.extent(0) ){
+    throw std::runtime_error("KokkosBlas: hermitian_matrix_vector_product: A.extent(1) != x.extent(0) ");
+  }
+  if ( A.extent(0) != y.extent(0) ){
+    throw std::runtime_error("KokkosBlas: hermitian_matrix_vector_product: A.extent(0) != y.extent(0) ");
+  }
+
+  // for non-complex ELementType_A, A can be treated as symmetric
```
mhoemmen, line 69:
```suggestion
  // for non-complex ElementType_A, A can be treated as symmetric
```

mhoemmen (approved):

PR #179, opened 2022-04-20
Make `conj_if_needed` work with `Kokkos::complex` #173

### LINKED ISSUE #173: Factor out `conj_if_needed` customization point
A lot of PRs, like https://github.com/kokkos/stdBLAS/pull/167 , repeat the definition of `conj_if_needed`.  Let's pull it out into a common header and make it a customization point (for e.g., `Kokkos::complex<T>` and potentially other user-defined complex types).

MikolajZuzek:
Fixes #173 by adding `Kokkos::complex<T>` support to `std::experimental::linalg::impl::conj_if_needed()` and removing replicated definitions.

### DIFF
--- a/tests/kokkos-based/hermitian_matrix_rank2_update_kokkos.cpp
+++ b/tests/kokkos-based/hermitian_matrix_rank2_update_kokkos.cpp
@@ -51,13 +51,13 @@ using namespace kokkostesting;
 template<class x_t, class y_t, class A_t, class Triangle>
 void hermitian_matrix_rank_2_update_gold_solution(const x_t &x, const y_t &y, A_t &A, Triangle /* t */)
 {
+  using std::experimental::linalg::impl::conj_if_needed;
   using size_type = std::experimental::extents<>::size_type;
-  constexpr auto conj = std::experimental::linalg::impl::conj_if_needed;
   constexpr bool low = std::is_same_v<Triangle, std::experimental::linalg::lower_triangle_t>;
   for (size_type j = 0; j < A.extent(1); ++j) {
     const size_type i1 = low ? A.extent(0) : j + 1;
     for (size_type i = low ? j : 0; i < i1; ++i) {
-      A(i,j) += x(i) * conj(y(j)) + y(i) * conj(x(j));
+      A(i,j) += x(i) * conj_if_needed(y(j)) + y(i) * conj_if_needed(x(j));
     }
   }
 }
--- a/tests/kokkos-based/hermitian_matrix_rank2_update_kokkos.cpp
+++ b/tests/kokkos-based/hermitian_matrix_rank2_update_kokkos.cpp
@@ -51,14 +51,13 @@ using namespace kokkostesting;
 template<class x_t, class y_t, class A_t, class Triangle>
 void hermitian_matrix_rank_2_update_gold_solution(const x_t &x, const y_t &y, A_t &A, Triangle /* t */)
 {
-  using KokkosKernelsSTD::mtxr2update_impl::conj_if_needed;
   using size_type = std::experimental::extents<>::size_type;
+  constexpr auto conj = std::experimental::linalg::impl::conj_if_needed;
   constexpr bool low = std::is_same_v<Triangle, std::experimental::linalg::lower_triangle_t>;
   for (size_type j = 0; j < A.extent(1); ++j) {
     const size_type i1 = low ? A.extent(0) : j + 1;
     for (size_type i = low ? j : 0; i < i1; ++i) {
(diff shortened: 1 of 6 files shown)


tests/kokkos-based/hermitian_matrix_rank2_update_kokkos.cpp
```diff
@@ -0,0 +1,330 @@
...
+#include "helpers.hpp"
+
+namespace{
+
+////////////////////////////////////////////////////////////
+// Utilities
+////////////////////////////////////////////////////////////
+
+// create rank-1 mdspan (vector)
+template <typename value_type,
```
mhoemmen, line 54: It seems like these functions are being repeated in every file.  Would you consider factoring them out into a separate header?  If so, then would you consider the following changes:

1. removing `inline`, as it is superfluous in this case; and
2. using the pointer type conversion function defined elsewhere (I think via PR #162) to `reinterpret_cast` `value_type` from `std::complex` to `Kokkos::complex` if needed?

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_matrix_rank_2_update.hpp
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  Impl::signal_kokkos_impl_called("hermitian_matrix_rank2_update");
+
+  // convert mdspans to views
+  const auto x_view = Impl::mdspan_to_view(x);
+  const auto y_view = Impl::mdspan_to_view(y);
+  auto A_view = Impl::mdspan_to_view(A);
+
+  constexpr auto conj = std::experimental::linalg::impl::conj_if_needed;
```
mhoemmen, line 275: [Not really optional]

Would you consider naming this differently?  Readers who miss line 275 may be tempted to "fix" line 279 by adding `std::` to the `conj` calls.  This may also cause shadowing warnings for users who (ill-advisedly) put `using std::conj;` in the global namespace.

mhoemmen (changes requested):

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_matrix_rank_2_update.hpp
MikolajZuzek: Thanks! Renamed to `conj_if_needed` on https://github.com/kokkos/stdBLAS/pull/179/commits/9448ebca5d7ad5c463dd39debfc882fd3e0f98ad.

tests/kokkos-based/hermitian_matrix_rank2_update_kokkos.cpp
MikolajZuzek: With pleasure - I hate duplication myself!
Please see PR https://github.com/kokkos/stdBLAS/pull/187 that reconciles test utilities and fixes `inline`.

MikolajZuzek: Ad.2. Where exactly would you see the cast to `Kokkos::complex` ?
Here `make_mdspan()` creates `mdspan` (which usually carries `std::complex`) and in the mapper we create `Kokkos::View` (which carries `Kokkos::complex`).

MikolajZuzek: Right, but that's `mdspan` to `View` conversion, so I'm missing out on where would you like me to cast to `Kokkos::complex` ? Here in the utilities we create and use `mdspan`s - except of [`parallel_reduce` in `is_same_matrix`](https://github.com/NexGenAnalytics/stdBLAS/blob/173_refactor_conj-if-needed/tests/kokkos-based/matrix_rank1_update_kokkos.cpp#L187-L196) which already uses `mdspan_to_view()` to prforming the conversions under consideration.

So where would you like this conversion and `Kokkos::complex` to appear ?
Your comment is attached to `make_mdspan()`, but it creates `mdspan` carrying `std::complex`, not `View` carrying `Kokkos::complex` ?

MikolajZuzek: Seems like this discussion is aimed at https://github.com/kokkos/stdBLAS/pull/187, which was already merged ?
@mhoemmen Please let me know if you'd wish any changes on this.

MikolajZuzek:
@mhoemmen @fnrizzi 
I just rebased this PR on current `main` (without any other changes).
Would you see anything else to be fixed on added here ?

mhoemmen (approved):

PR #180, opened 2022-04-21
[ready] kokkos impl of `matrix_{frob,one,inf}_norm` and `vector_sum_of_squares`

fnrizzi:


tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_matrix_frob_norm_kk.hpp
```diff
@@ -0,0 +1,53 @@
...
+  if (A.extent(0) == zero || A.extent(1) == zero) {
+    return init;
+  }
+  else if(A.extent(0) == one && A.extent(1) == one) {
+    using std::abs;
+    return init + abs(A(0, 0));
+  }
+
+  const std::size_t totNumElements = A.extent(0)*A.extent(1);
+  using flatten_t = std::experimental::mdspan<
```
mhoemmen, line 39: [Required] This will only work correctly for `LayoutLeft` and `LayoutRight`, not for `LayoutStride`.  (In fact, historically, `LayoutLeft` and `LayoutRight` would introduce padding.  I think that's gone now, though.)  Would you consider at least constraining `Layout` to be either `LayoutLeft` or `LayoutRight`?  A fall-back implementation could just use a Kokkos mirror view.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_matrix_inf_norm_kk.hpp
```diff
@@ -0,0 +1,52 @@
...
+  if (A.extent(0) == 0){
+    return init;
+  }
+
+  auto A_view = Impl::mdspan_to_view(A);
+
+  Scalar result = {};
+  Kokkos::Max<Scalar> reducer(result);
+  Kokkos::parallel_reduce(Kokkos::RangePolicy(ExeSpace(), 0, A_view.extent(0)),
+			  KOKKOS_LAMBDA (const std::size_t i, Scalar & update)
```
mhoemmen, line 36: [Optional] Would you consider one of the following options?

1. Using `Kokkos::MDRange`,
2. using nested parallelism, or
3. flattening out the parallelism?

For example, Option (3) with iterating over rows would look like this.

```c++
Kokkos::parallel_reduce(
  Kokkos::RangePolicy(ExeSpace(), 0, A_view.extent(0) * A_view.extent(1)),
  KOKKOS_LAMBDA (const std::size_t k, Scalar& update) {
    using ats = Kokkos::Details::ArithTraits<ElementType>;
    const auto i = k / A_view.extent(1);
    const auto j = k % A_view.extent(1);
    reducer.join(update, ats::abs(A_view(i, j)));
  }, reducer);
```

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_vector_sum_of_squares_kk.hpp
```diff
@@ -2,21 +2,66 @@
...
-  std::cout << "vector_sum_of_squares: kkernels impl missing !!!\n";
-  return init;
+
+  Impl::signal_kokkos_impl_called("vector_sum_of_squares");
+
+  auto x_view = Impl::mdspan_to_view(x);
+  std::experimental::linalg::sum_of_squares_result<Scalar> result;
+
+  // FRIZZI:
+  // (1) this works only for types that have a reduction identity.
```
mhoemmen, line 30: Just use `T{}` for the reduction identity.  We may need some additional wording in P1673 for the parallel algorithms.  I've filed a wording issue for that: https://github.com/kokkos/stdBLAS/issues/188

mhoemmen (changes requested):

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_matrix_frob_norm_kk.hpp
fnrizzi: oh yes, you are right, thanks for catching this! i will fix it.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_matrix_inf_norm_kk.hpp
fnrizzi: yes definitely! My plan was to start with a "first" impl of every function and then go back to improve those impl.

fnrizzi: is it ok if I make these improvements in a followup PR?

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_matrix_frob_norm_kk.hpp
fnrizzi: ok, I think I fixed it. I am basically doing the scaled sum of square approach but without assuming any layout.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_matrix_inf_norm_kk.hpp
fnrizzi: done for this!

PR #181, opened 2022-04-21
[ready] kokkos impl: `copy`, `idx_abs_max`

fnrizzi:


tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_copy_kk.hpp
```diff
@@ -0,0 +1,59 @@
...
+	    std::experimental::extents<ext_y ...>,
+	    Layout_y,
+	    std::experimental::default_accessor<ElementType_y>
+	  > y)
+requires ((sizeof...(ext_x) == sizeof...(ext_y)) && (x.rank() <=2))
+{
+  Impl::signal_kokkos_impl_called("copy");
+
+  auto x_view = Impl::mdspan_to_view(x);
+  auto y_view = Impl::mdspan_to_view(y);
```
mhoemmen, line 34: You have the `Kokkos::View` and an execution space, so would you consider just using `Kokkos::deep_copy` here?

mhoemmen (changes requested):

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_copy_kk.hpp
fnrizzi: I think that if the layout of x and y are different, the deep copy fails. To use deep copy I would need to specialize for when the layouts are the same. Right?

fnrizzi: done!

mhoemmen:
Thanks so much!

PR #184, opened 2022-04-21
layout_blas_general: Add typename to fix C++ < 20

mhoemmen:
@youyu3 reports that when building with C++ < 20, this file is the first to report build errors:

```
".../include/experimental/__p1673_bits/layout_blas_general.hpp", line 154: error: nontype "std::experimental::__p1673_version_0::linalg::__layout_blas_general_impl::__layout_blas_impl<BaseLayout, StaticLDA>::__extents_type::size_type" is not a type name
constexpr __extents_type::size_type required_span_size() const noexcept {
```
One possible cause is that C++20 relaxed where `typename` is needed.  This PR aims to fix that.

### DIFF
--- a/include/experimental/__p1673_bits/layout_blas_general.hpp
+++ b/include/experimental/__p1673_bits/layout_blas_general.hpp
@@ -151,22 +151,22 @@ public:
   MDSPAN_INLINE_FUNCTION constexpr __extents_type extents() const noexcept { return _base_layout.extents(); }
 
   MDSPAN_INLINE_FUNCTION
-  constexpr __extents_type::size_type required_span_size() const noexcept {
+  constexpr typename __extents_type::size_type required_span_size() const noexcept {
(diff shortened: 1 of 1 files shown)


mhoemmen:
FYI, there are other build errors if not enabling C++20 support.  This just aims to fix the easy ones : - ) .

youyu3:
Confirmed that these changes fixed all compilation errors in this file without C++20.

fnrizzi (approved):

MikolajZuzek (approved):
Thanks @mhoemmen: this issue and fix is consistent with what I've encountered while playing with CUDA (C++17) build.

mhoemmen:
@MikolajZuzek Thanks for the review! : - D  If you feel like making the C++17 build work, please don't be shy to fix whatever you encounter as you're making other changes! : - )  I'll try to keep PRs like this small and self-contained, so they don't interfere with your work.

PR #187, opened 2022-04-22
kokkos: refactor duplicated test helpers

MikolajZuzek:
This PR moves common Kokkos test utilities to shared headers - refactoring massive duplication and significantly shortening test files.


tests/kokkos-based/gtest_fixtures.hpp
```diff
@@ -326,4 +326,10 @@ using blas2_signed_float_fixture  = _blas2_signed_fixture<float>;
 using blas2_signed_double_fixture = _blas2_signed_fixture<double>;
 using blas2_signed_complex_double_fixture = _blas2_signed_fixture<std::complex<double>>;
 
+// Run all tests for types defined above
+#define FOR_ALL_BLAS2_TYPES(TEST_DEF) \
+  TEST_DEF(double) \
+  TEST_DEF(float) \
+  TEST_DEF(complex_double)
```
mhoemmen, line 333: Awww, no `complex_float`?

tests/kokkos-based/helpers.hpp
```diff
@@ -84,5 +84,235 @@ auto create_stdvector_and_copy_rowwise(T sourceView)
...
 }
 
+// create rank-1 mdspan (vector)
+template <typename ValueType,
+          typename mdspan_t = typename _blas2_signed_fixture<ValueType>::mdspan_r1_t>
+mdspan_t make_mdspan(ValueType *data, std::size_t ext) {
+  return mdspan_t(data, ext);
+}
+
+template <typename ValueType,
```
mhoemmen, line 94: I don't think you actually need the `const ValueType` overload, as the `ValueType` generic version should deduce `const` as needed.  For example, the following should deduce `const ValueType` as the `element_type` of the `mdspan`.

```c++
template <typename ValueType>
auto make_mdspan(const std::vector<ValueType> &v) {
  return make_mdspan(v.data(), v.size()); // v.data() returns const ValueType*
}
```

tests/kokkos-based/helpers.hpp
```diff
@@ -84,5 +84,235 @@ auto create_stdvector_and_copy_rowwise(T sourceView)
...
+// no-tolerance (exact) comparison
+// TODO: add tolerance based comparison if needed (see is_same_matrix below)
+template <typename ElementType1,
+          typename LayoutPolicy1,
+          typename AccessorPolicy1,
+          typename ElementType2,
+          typename LayoutPolicy2,
+          typename AccessorPolicy2>
+bool is_same_vector(
+    const mdspan<ElementType1, extents<dynamic_extent>, LayoutPolicy1, AccessorPolicy1> &v1,
```
mhoemmen, line 126: The Standard Library convention for view types like `mdspan`, `span`, and `string_view` is to pass them by value, not by const reference.

tests/kokkos-based/helpers.hpp
```diff
@@ -84,5 +84,235 @@ auto create_stdvector_and_copy_rowwise(T sourceView)
...
+{
+  const auto size = v1.extent(0);
+  if (size != v2.extent(0))
+    return false;
+  const auto v1_view = KokkosKernelsSTD::Impl::mdspan_to_view(v1);
+  const auto v2_view = KokkosKernelsSTD::Impl::mdspan_to_view(v2);
+  int diff = false;
+  Kokkos::parallel_reduce(size,
+    KOKKOS_LAMBDA(const std::size_t i, decltype(diff) &d){
+        d = d || !(v1_view(i) == v2_view(i));
```
mhoemmen, line 137: Would you consider using the `LOr` reducer to document intent, instead of rolling the reduction by hand?  (The element access function would then just be `v1_view(i) != v2_view(i)`.)

https://github.com/kokkos/kokkos/blob/2834f94af9b01debf67c1aaa3f0eb0c903d72c8d/core/src/Kokkos_Parallel_Reduce.hpp#L301

Also, would you consider using `operator[]` instead of `operator()` to access the view?  That works now, and would be more consistent with C++23.

tests/kokkos-based/helpers.hpp
```diff
@@ -84,5 +84,235 @@ auto create_stdvector_and_copy_rowwise(T sourceView)
...
+    const std::vector<ElementType> &v2)
+{
+  return is_same_vector(make_mdspan(v1), make_mdspan(v2));
+}
+
+// real diff: d = |v1 - v2|
+template <typename T, typename Enabled=void>
+class value_diff {
+public:
+  value_diff(const T &val1, const T &val2): _v(fabs(val1 - val2)) {}
```
mhoemmen, line 176: Would you consider using `std::abs` instead of `fabs`, so that integer arguments aren't narrowed to floating-point values?

tests/kokkos-based/helpers.hpp
```diff
@@ -84,5 +84,235 @@ auto create_stdvector_and_copy_rowwise(T sourceView)
...
+  operator T() const { return _v; }
+protected:
+  value_diff() = default;
+  T _v;
+};
+
+// real diff: d = max(|R(v1) - R(v2)|, |I(v1) - I(v2)|)
+// Note: returned value is of underlying real type
+template <typename T>
+class value_diff<std::complex<T>>: public value_diff<T> {
```
mhoemmen, line 186: [Optionally,] would you consider the following class hierarchy changes?

1. Make `_v` private.
2. Initialize `base` in the constructor, like this:

```c++
  value_diff(const std::complex<T> &val1, const std::complex<T> &val2) : base(base(val1.real(), val2.real()), base(val1.imag(), val2.imag())) {}
    const T dimag = base(val1.imag(), val2.imag());
    base::_v = std::max(dreal, dimag);
  }
```

I'm not a huge fan of `value_diff<complex<T>>` inheriting from `value_diff<T>`, but it seems to be working.

mhoemmen (approved):

tests/kokkos-based/gtest_fixtures.hpp
fnrizzi: That is my fault because I started all tests using `double`, `float` and `complex<double>`, so Miko was following my setup :) 
I created a new issue: https://github.com/kokkos/stdBLAS/issues/190 to add `complex<float>` later

tests/kokkos-based/helpers.hpp
```diff
@@ -84,5 +84,235 @@ auto create_stdvector_and_copy_rowwise(T sourceView)
...
+  // compute gold
+  auto A_copy = create_stdvector_and_copy_rowwise(A);
+  auto A_gold = make_mdspan(A_copy.data(), A.extent(0), A.extent(1));
+  get_gold(A_gold);
+
+  // run tested routine
+  action();
+
+  // compare results with gold
+  EXPECT_TRUE(is_same_matrix(A_gold, A, tolerance<value_type>(1e-9, 1e-2f)));
```
fnrizzi, line 301: what if we have to use different tolerances for different algorithms?

tests/kokkos-based/helpers.hpp
```diff
@@ -84,5 +84,235 @@ auto create_stdvector_and_copy_rowwise(T sourceView)
...
+
+template <> double tolerance<double>(double double_tol, float float_tol) { return double_tol; }
+template <> float  tolerance<float>( double double_tol, float float_tol) { return float_tol; }
+template <> double tolerance<std::complex<double>>(double double_tol, float float_tol) { return double_tol; }
+template <> float  tolerance<std::complex<float>>( double double_tol, float float_tol) { return float_tol; }
+
+// checks if std::complex<T> and Kokkos::complex<T> are aligned
+// (they can get misalligned when Kokkos is build with Kokkos_ENABLE_COMPLEX_ALIGN=ON)
+template <typename ValueType, typename Enabled = void>
+struct check_types: public std::true_type {};
```
fnrizzi, line 256: even if internal, I would rename `check_types` -> `check_complex_alignment` or something similar

tests/kokkos-based/gtest_fixtures.hpp
fnrizzi: @MikolajZuzek my preference would be to remove this macro. It is just 3 lines of code :) it does not save you much typing. Also, I think it adds another layer that one would have to find when looking at the tests. 
I think this gtest_fixtures should only contain the fixtures. If you really want this macro, maybe we can put it in another file called `test_macros.hpp` or similar so that it is clear the distinction.

tests/kokkos-based/helpers.hpp
MikolajZuzek: Correct - thank you for pointing this out! Fixed on https://github.com/kokkos/stdBLAS/pull/187/commits/2f72e982654431235dcd7bd0f6c2f786954ed329.

tests/kokkos-based/helpers.hpp
MikolajZuzek: Fixed on https://github.com/kokkos/stdBLAS/pull/187/commits/c0c4e70117366fc32f7d8f41bccff80077610f9c.

tests/kokkos-based/helpers.hpp
MikolajZuzek: Fixed on https://github.com/kokkos/stdBLAS/pull/187/commits/e290d12cd00367cd6a399030c52e274cd8c2cd98.

tests/kokkos-based/helpers.hpp
MikolajZuzek: Can I replace whole inheritance with much simpler and cleaner function:
```cpp
template <typename RealValue>
RealValue scalar_diff(RealValue v1, RealValue v2) {
  return std::abs(v2 - v1);
}

template <typename T>
T scalar_diff(const std::complex<T> &v1, const std::complex<T> &v2) {
  const auto dr = scalar_diff(v1.real(), v2.real());
  const auto di = scalar_diff(v1.imag(), v2.imag());
  return std::max(dr, di);
}

template <typename T>
T scalar_diff(const Kokkos::complex<T> &v1, const Kokkos::complex<T> &v2) {
  const auto dr = scalar_diff(v1.real(), v2.real());
  const auto di = scalar_diff(v1.imag(), v2.imag());
  return dr > di ? dr : di; // can't use std::max on GPU
}
```
?

mhoemmen: YES PLEASE : - )

Also, you might need to bless all but the `std::complex` overloads with `KOKKOS_INLINE_FUNCTION`, no?  Things might be different now, though.

crtrott (approved):
Certainly a good improvement. But where is this _blas2_signed_fixture thing coming from? Is that some semi pulic non-namespaced thing in our headers?

PR #191, opened 2022-04-26
transposed: Add typename to fix C++ < 20

youyu3:
Same change as in PR #184.


mhoemmen (approved):
Thanks for the fix!  I'm not convinced that C++20 improved the rules for when to use `typename` ; - ) .

PR #192, opened 2022-04-26
Misc changes to make it work for C++17

youyu3:
* Replace `requires` with the mdspan macro `MDSPAN_TEMPLATE_REQUIRES`, which
  uses `enable_if_t` when C++20 is not enabled
* Replace `template<std::floating_point T>` with a `requires`/`enable_if_t`
  (via above mdspan macro)

### DIFF
--- a/include/experimental/__p1673_bits/blas1_givens.hpp
+++ b/include/experimental/__p1673_bits/blas1_givens.hpp
@@ -45,7 +45,6 @@
 
 #include <cmath>
 #include <complex>
-#include <concepts>
 
 namespace std {
 namespace experimental {
@@ -112,7 +111,7 @@ struct is_custom_givens_rotation_apply_avail<
 
 
 
-template<std::floating_point Real>
+MDSPAN_TEMPLATE_REQUIRES( class Real, /* requires */ ( _MDSPAN_TRAIT(is_floating_point, Real) ) )
 void givens_rotation_setup(const Real f,
                            const Real g,
                            Real& cs,
@@ -218,7 +217,7 @@ void givens_rotation_setup(const Real f,
 }
 
 namespace impl {
-template<std::floating_point Real>
+MDSPAN_TEMPLATE_REQUIRES( class Real, /* requires */ ( _MDSPAN_TRAIT(is_floating_point, Real) ) )
 Real abs1(const complex<Real>& ff) {
   using std::abs;
   using std::imag;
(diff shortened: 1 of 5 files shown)


include/experimental/__p1673_bits/blas1_linalg_swap.hpp
```diff
@@ -127,15 +127,17 @@ struct is_custom_vector_swap_elements_avail<
...
+         class ElementType_x,
          class Layout_x,
          class Accessor_x,
          class ElementType_y,
-         extents<>::size_type ... ext_y,
          class Layout_y,
-         class Accessor_y>
-  requires (sizeof...(ext_x) == sizeof...(ext_y))
+         class Accessor_y,
+         extents<>::size_type ... ext_x,
```
mhoemmen, line 137: @youyu3 Just curious why you changed the order of template parameters here and in a few other places -- did the macro not work otherwise?  Would it help to surround the macro argument with parentheses, like this: `(extents<>::size_type ... ext_x)`

mhoemmen (approved):

include/experimental/__p1673_bits/blas1_linalg_swap.hpp
youyu3: Aha, with C++17, it didn't compile before I switched to the macro, with an error that says parameter packs should be at the end of the parameter list.  I just tested and confirmed that the original order works with the macro.  I'll back these out.

youyu3: Backed them out.

mhoemmen:
Thanks so much! : - )

include/experimental/__p1673_bits/blas1_givens.hpp
```diff
@@ -372,16 +371,18 @@ void givens_rotation_setup(const complex<Real>& f,
...
          extents<>::size_type ext1,
          class Layout1,
          class Accessor1,
          class ElementType2,
          extents<>::size_type ext2,
          class Layout2,
          class Accessor2,
-         std::floating_point Real>
+         class Real,
+         /* requires */ (_MDSPAN_TRAIT(is_floating_point, Real))
```
MikolajZuzek, line 384: @youyu3 g++11 on Ubuntu 20.04 fails here with:
```shell
include/experimental/__p1673_bits/blas1_givens.hpp:384:25: error: expected identifier before ‘(’ token
  384 |          /* requires */ (_MDSPAN_TRAIT(is_floating_point, Real))
      |                         ^
```
(cases with only `Real` in `MDSPAN_TEMPLATE_REQUIRES` compile fine)

Is it something on my side ?

MikolajZuzek: Updating `mdspan` solved the problem, I'm sorry for bothering.

youyu3: Thanks for the update!  Should've noted that this needs updated `mdspan`.

PR #193, opened 2022-04-30
kokkos: change tolerace-based comparison to diff and add relative difference

MikolajZuzek:
On this PR I'd like to propose:
* changing tolerance based `is_same_{vector,matrix}` comparison to `{vector,matrix}_abs_diff` difference calculation which can be used like `EXPECT_LE(matrix_abs_diff(A_gold, A), A_tolerance)` in tests and show compared numerical values on failure;
* adding new `{vector,matrix}_rel_diff` variant that computes _relative_ difference;
* revision of tolerance values in existing tests;

### DIFF
--- a/tests/kokkos-based/helpers.hpp
+++ b/tests/kokkos-based/helpers.hpp
@@ -140,14 +140,16 @@ auto scalar_ref_diff(const T &v1, const T &v2) {
 
 // no-tolerance (exact) comparison
 template <typename ElementType1,
+          std::size_t Extent1,
           typename LayoutPolicy1,
           typename AccessorPolicy1,
           typename ElementType2,
+          std::size_t Extent2,
           typename LayoutPolicy2,
           typename AccessorPolicy2>
 bool is_same_vector(
-    mdspan<ElementType1, extents<dynamic_extent>, LayoutPolicy1, AccessorPolicy1> v1,
-    mdspan<ElementType2, extents<dynamic_extent>, LayoutPolicy2, AccessorPolicy2> v2)
+    mdspan<ElementType1, extents<Extent1>, LayoutPolicy1, AccessorPolicy1> v1,
+    mdspan<ElementType2, extents<Extent2>, LayoutPolicy2, AccessorPolicy2> v2)
 {
   const auto size = v1.extent(0);
   if (size != v2.extent(0))
@@ -166,23 +168,25 @@ bool is_same_vector(
 }
 
 template <typename ElementType1,
+          std::size_t Extent,
           typename LayoutPolicy,
           typename AccessorPolicy,
           typename ElementType2>
 bool is_same_vector(
-    mdspan<ElementType1, extents<dynamic_extent>, LayoutPolicy, AccessorPolicy> v1,
+    mdspan<ElementType1, extents<Extent>, LayoutPolicy, AccessorPolicy> v1,
     const std::vector<ElementType2> &v2)
 {
   return is_same_vector(v1, make_mdspan(v2));
 }
 
 template <typename ElementType1,
+          std::size_t Extent,
           typename LayoutPolicy,
           typename AccessorPolicy,
           typename ElementType2>
 bool is_same_vector(
     const std::vector<ElementType1> &v1,
-    mdspan<ElementType2, extents<dynamic_extent>, LayoutPolicy, AccessorPolicy> v2)
+    mdspan<ElementType2, extents<Extent>, LayoutPolicy, AccessorPolicy> v2)
 {
   return is_same_vector(v2, v1);
 }
@@ -196,14 +200,16 @@ bool is_same_vector(
 }
 
 template <typename ElementType1,
+          std::size_t Extent1,
           typename LayoutPolicy1,
           typename AccessorPolicy1,
           typename ElementType2,
+          std::size_t Extent2,
           typename LayoutPolicy2,
           typename AccessorPolicy2>
 auto vector_abs_diff(
-    mdspan<ElementType1, extents<dynamic_extent>, LayoutPolicy1, AccessorPolicy1> v1,
-    mdspan<ElementType2, extents<dynamic_extent>, LayoutPolicy2, AccessorPolicy2> v2)
+    mdspan<ElementType1, extents<Extent1>, LayoutPolicy1, AccessorPolicy1> v1,
+    mdspan<ElementType2, extents<Extent2>, LayoutPolicy2, AccessorPolicy2> v2)
 {
   using RetType = decltype(scalar_abs_diff(v1[0], v2[0])); // will be same for views
   const auto size = v1.extent(0);
@@ -223,23 +229,25 @@ auto vector_abs_diff(
 }
 
 template <typename ElementType1,
+          std::size_t Extent,
           typename LayoutPolicy,
           typename AccessorPolicy,
           typename ElementType2>
 auto vector_abs_diff(
-    mdspan<ElementType1, extents<dynamic_extent>, LayoutPolicy, AccessorPolicy> v1,
+    mdspan<ElementType1, extents<Extent>, LayoutPolicy, AccessorPolicy> v1,
     const std::vector<ElementType2> &v2)
 {
   return vector_abs_diff(v1, make_mdspan(v2));
 }
 
 template <typename ElementType1,
+          std::size_t Extent,
           typename LayoutPolicy,
           typename AccessorPolicy,
           typename ElementType2>
 auto vector_abs_diff(
     const std::vector<ElementType1> &v1,
-    mdspan<ElementType2, extents<dynamic_extent>, LayoutPolicy, AccessorPolicy> v2)
+    mdspan<ElementType2, extents<Extent>, LayoutPolicy, AccessorPolicy> v2)
 {
   return vector_abs_diff(v2, v1);
 }
@@ -253,14 +261,16 @@ auto vector_abs_diff(
 }
 
 template <typename ElementType1,
+          std::size_t Extent1,
           typename LayoutPolicy1,
           typename AccessorPolicy1,
           typename ElementType2,
+          std::size_t Extent2,
           typename LayoutPolicy2,
           typename AccessorPolicy2>
 auto vector_rel_diff(
-    mdspan<ElementType1, extents<dynamic_extent>, LayoutPolicy1, AccessorPolicy1> v1,
-    mdspan<ElementType2, extents<dynamic_extent>, LayoutPolicy2, AccessorPolicy2> v2)
+    mdspan<ElementType1, extents<Extent1>, LayoutPolicy1, AccessorPolicy1> v1,
+    mdspan<ElementType2, extents<Extent2>, LayoutPolicy2, AccessorPolicy2> v2)
 {
   using RetType = decltype(scalar_abs_diff(v1[0], v2[0]));
   const auto size = v1.extent(0);
@@ -286,23 +296,25 @@ auto vector_rel_diff(
 }
 
 template <typename ElementType1,
+          std::size_t Extent1,
           typename LayoutPolicy,
           typename AccessorPolicy,
           typename ElementType2>
 auto vector_rel_diff(
-    mdspan<ElementType1, extents<dynamic_extent>, LayoutPolicy, AccessorPolicy> v1,
+    mdspan<ElementType1, extents<Extent1>, LayoutPolicy, AccessorPolicy> v1,
     const std::vector<ElementType2> &v2)
 {
   return vector_rel_diff(v1, make_mdspan(v2));
 }
 
 template <typename ElementType1,
+          std::size_t Extent,
           typename LayoutPolicy,
           typename AccessorPolicy,
           typename ElementType2>
 auto vector_rel_diff(
     const std::vector<ElementType1> &v1,
-    mdspan<ElementType2, extents<dynamic_extent>, LayoutPolicy, AccessorPolicy> v2)
+    mdspan<ElementType2, extents<Extent>, LayoutPolicy, AccessorPolicy> v2)
 {
   return vector_rel_diff(v2, v1);
 }
@@ -317,14 +329,18 @@ auto vector_rel_diff(
 
 // no-tolerance (exact) comparison
 template <typename ElementType1,
+          std::size_t Extent10,
+          std::size_t Extent11,
           typename LayoutPolicy1,
           typename AccessorPolicy1,
           typename ElementType2,
+          std::size_t Extent20,
+          std::size_t Extent21,
           typename LayoutPolicy2,
           typename AccessorPolicy2>
 bool is_same_matrix(
-    mdspan<ElementType1, extents<dynamic_extent, dynamic_extent>, LayoutPolicy1, AccessorPolicy1> A,
-    mdspan<ElementType2, extents<dynamic_extent, dynamic_extent>, LayoutPolicy2, AccessorPolicy2> B)
+    mdspan<ElementType1, extents<Extent10, Extent11>, LayoutPolicy1, AccessorPolicy1> A,
+    mdspan<ElementType2, extents<Extent20, Extent21>, LayoutPolicy2, AccessorPolicy2> B)
 {
   const auto ext0 = A.extent(0);
   const auto ext1 = A.extent(1);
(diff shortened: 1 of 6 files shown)


tests/kokkos-based/helpers.hpp
```diff
@@ -166,60 +195,216 @@ bool is_same_vector(
...
+{
+  using RetType = decltype(scalar_abs_diff(v1[0], v2[0])); // will be same for views
+  const auto size = v1.extent(0);
+  if (size != v2.extent(0))
+    return std::numeric_limits<RetType>::max(); // very, very different
+  const auto v1_view = KokkosKernelsSTD::Impl::mdspan_to_view(v1);
+  const auto v2_view = KokkosKernelsSTD::Impl::mdspan_to_view(v2);
+  RetType diff = static_cast<RetType>(0);
+  Kokkos::parallel_reduce(size,
+    KOKKOS_LAMBDA(const std::size_t i, RetType &diff){
```
mhoemmen, line 216: Would you consider giving the inner `diff` variable a different name, to prevent shadowing warnings and make the code more clear?

tests/kokkos-based/helpers.hpp
```diff
@@ -166,60 +195,216 @@ bool is_same_vector(
...
-T scalar_diff(const std::complex<T> &v1, const std::complex<T> &v2) {
-  const auto dr = scalar_diff(v1.real(), v2.real());
-  const auto di = scalar_diff(v1.imag(), v2.imag());
-  return std::max(dr, di);
+template <typename ElementType1,
+          typename LayoutPolicy,
+          typename AccessorPolicy,
+          typename ElementType2>
+auto vector_abs_diff(
+    mdspan<ElementType1, extents<dynamic_extent>, LayoutPolicy, AccessorPolicy> v1,
```
mhoemmen, line 230: Using `dynamic_extent` is fine here, but would you consider instead templating on the extent, like the following?
```c++
template <typename ElementType1,
          std::size_t Extent,
          typename LayoutPolicy,
          typename AccessorPolicy,
          typename ElementType2>
auto vector_abs_diff(
    mdspan<ElementType1, extents<Extent>, LayoutPolicy, AccessorPolicy> v1,
    const std::vector<ElementType2>& v2) { /* ... */ }
```
That would make it possible to compare a `vector` with an `mdspan` of compile-time extent.

tests/kokkos-based/helpers.hpp
```diff
@@ -166,60 +195,216 @@ bool is_same_vector(
...
+{
+  using RetType = decltype(scalar_abs_diff(v1[0], v2[0]));
+  const auto size = v1.extent(0);
+  if (size != v2.extent(0))
+    return std::numeric_limits<RetType>::max(); // very, very different
+
+  constexpr auto zero1 = static_cast<ElementType1>(0);
+  constexpr auto zero2 = static_cast<ElementType2>(0);
+  RetType abs_diff = vector_abs_diff(v1, v2);
+  auto v1_norm = scalar_abs_diff(zero1, v1[std::experimental::linalg::idx_abs_max(v1)]);
```
mhoemmen, line 273: Would you consider testing the index first, before using it to index into a vector?

tests/kokkos-based/helpers.hpp
```diff
@@ -166,60 +195,216 @@ bool is_same_vector(
...
+          typename LayoutPolicy2,
+          typename AccessorPolicy2>
+auto vector_rel_diff(
+    mdspan<ElementType1, extents<dynamic_extent>, LayoutPolicy1, AccessorPolicy1> v1,
+    mdspan<ElementType2, extents<dynamic_extent>, LayoutPolicy2, AccessorPolicy2> v2)
+{
+  using RetType = decltype(scalar_abs_diff(v1[0], v2[0]));
+  const auto size = v1.extent(0);
+  if (size != v2.extent(0))
+    return std::numeric_limits<RetType>::max(); // very, very different
```
mhoemmen, line 268: Please see above note; thanks!

tests/kokkos-based/helpers.hpp
```diff
@@ -166,60 +195,216 @@ bool is_same_vector(
...
+    mdspan<ElementType2, extents<dynamic_extent, dynamic_extent>, LayoutPolicy2, AccessorPolicy2> B)
 {
   const auto ext0 = A.extent(0);
   const auto ext1 = A.extent(1);
   if (B.extent(0) != ext0 or B.extent(1) != ext1)
     return false;
   const auto A_view = KokkosKernelsSTD::Impl::mdspan_to_view(A);
   const auto B_view = KokkosKernelsSTD::Impl::mdspan_to_view(B);
-  // Note: reducint to `int` because Kokkos can complain on `bool` not being 
+  // Note: reducint to `int` because Kokkos can complain on `bool` not being
```
mhoemmen, line 335:
```suggestion
  // Note: reducing to `int` because Kokkos can complain on `bool` not being
```

tests/kokkos-based/helpers.hpp
```diff
@@ -166,60 +195,216 @@ bool is_same_vector(
...
+          typename AccessorPolicy2>
+auto matrix_abs_diff(
+    mdspan<ElementType, extents<dynamic_extent, dynamic_extent>, LayoutPolicy1, AccessorPolicy1> A,
+    mdspan<ElementType, extents<dynamic_extent, dynamic_extent>, LayoutPolicy2, AccessorPolicy2> B)
+{
+  const auto ext0 = A.extent(0);
+  const auto ext1 = A.extent(1);
+  using RetType = decltype(scalar_abs_diff(A(0, 0), B(0, 0)));
+  if (B.extent(0) != ext0 or B.extent(1) != ext1) {
+    return std::numeric_limits<RetType>::max(); // very, very different
```
mhoemmen, line 381: Please see above note; thanks!

tests/kokkos-based/helpers.hpp
```diff
@@ -166,60 +195,216 @@ bool is_same_vector(
...
+  }
+  return vector_abs_diff(
```
mhoemmen, line 383: [REQUIRED]

This is incorrect for any layout other than `layout_left` or `layout_right`.  In particular, this may give incorrect results for any `layout_stride` matrix with leading dimension greater than the number of rows (a use case that the C and Fortran BLAS both support).

There are two approaches for fixing this.

1. Constrain _both_ layouts to be either `layout_left` or `layout_right`.
2. Write a generic implementation.

tests/kokkos-based/helpers.hpp
```diff
@@ -166,60 +195,216 @@ bool is_same_vector(
...
+    mdspan<ElementType, extents<dynamic_extent, dynamic_extent>, LayoutPolicy1, AccessorPolicy1> A,
+    mdspan<ElementType, extents<dynamic_extent, dynamic_extent>, LayoutPolicy2, AccessorPolicy2> B)
+{
+  const auto ext0 = A.extent(0);
+  const auto ext1 = A.extent(1);
+  using RetType = decltype(scalar_abs_diff(A(0, 0), B(0, 0)));
+  if (B.extent(0) != ext0 or B.extent(1) != ext1) {
+    return std::numeric_limits<RetType>::max(); // very, very different
+  }
+  return vector_rel_diff(
```
mhoemmen, line 403: Please see above note; thanks!

mhoemmen (changes requested):
Thanks for working on this! : - )

tests/kokkos-based/helpers.hpp
```diff
@@ -108,8 +108,37 @@ mdspan_t make_mdspan(ValueType *data, std::size_t ext0, std::size_t ext1) {
...
+
+template <typename T>
+KOKKOS_INLINE_FUNCTION
+T scalar_abs_diff(const Kokkos::complex<T> &v1, const Kokkos::complex<T> &v2) {
+  const auto dr = scalar_abs_diff(v1.real(), v2.real());
+  const auto di = scalar_abs_diff(v1.imag(), v2.imag());
+  return dr > di ? dr : di; // can't use std::max on GPU
+}
+
+#if 0
```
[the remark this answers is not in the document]
MikolajZuzek: Fixed on https://github.com/kokkos/stdBLAS/pull/193/commits/5b74a1ea22693401d5997ac5562d9b1ce7512949.

tests/kokkos-based/helpers.hpp
MikolajZuzek: Fixed on https://github.com/kokkos/stdBLAS/pull/193/commits/3ff22035ce62ddbd03314185ef945301d7228444.

tests/kokkos-based/helpers.hpp
MikolajZuzek: Thank you for pointing this out! Fixed on https://github.com/kokkos/stdBLAS/pull/193/commits/860d497d1f77de874de43e69af7bfea608e23e2b.

tests/kokkos-based/helpers.hpp
```diff
@@ -166,60 +195,216 @@ bool is_same_vector(
...
+          typename LayoutPolicy2,
+          typename AccessorPolicy2>
+auto vector_abs_diff(
+    mdspan<ElementType1, extents<dynamic_extent>, LayoutPolicy1, AccessorPolicy1> v1,
+    mdspan<ElementType2, extents<dynamic_extent>, LayoutPolicy2, AccessorPolicy2> v2)
+{
+  using RetType = decltype(scalar_abs_diff(v1[0], v2[0])); // will be same for views
+  const auto size = v1.extent(0);
+  if (size != v2.extent(0))
+    return std::numeric_limits<RetType>::max(); // very, very different
```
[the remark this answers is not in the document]
MikolajZuzek: Thank you, I was not sure which approach is better. Fixed on https://github.com/kokkos/stdBLAS/pull/193/commits/72dd1f02ff0a0bad7dc88649df2a1dec7a3e8b20.

tests/kokkos-based/helpers.hpp
MikolajZuzek: Also fixed with https://github.com/kokkos/stdBLAS/pull/193/commits/72dd1f02ff0a0bad7dc88649df2a1dec7a3e8b20.

tests/kokkos-based/helpers.hpp
MikolajZuzek: Also fixed with https://github.com/kokkos/stdBLAS/pull/193/commits/72dd1f02ff0a0bad7dc88649df2a1dec7a3e8b20.

tests/kokkos-based/helpers.hpp
MikolajZuzek: Thank you for pointing this out! Fixed with generic implementation on https://github.com/kokkos/stdBLAS/pull/193/commits/a945906066e35af077cbdcbf0cfdccb55ea2eeb3.

tests/kokkos-based/helpers.hpp
MikolajZuzek: Fixed as [noted above](https://github.com/kokkos/stdBLAS/pull/193#discussion_r862937205).

tests/kokkos-based/helpers.hpp
```diff
@@ -108,8 +108,37 @@ mdspan_t make_mdspan(ValueType *data, std::size_t ext0, std::size_t ext1) {
   return mdspan_t(data, ext0, ext1);
 }
 
+template <typename RealValue>
+KOKKOS_INLINE_FUNCTION
+RealValue scalar_abs_diff(RealValue v1, RealValue v2) {
+  return std::abs(v2 - v1);
```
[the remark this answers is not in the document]
MikolajZuzek: Thanks! Fixed on https://github.com/kokkos/stdBLAS/pull/193/commits/b1622086e153e0e872fc2605fa9527817fa15102.

tests/kokkos-based/helpers.hpp
MikolajZuzek: I'm sorry... Fixed on https://github.com/kokkos/stdBLAS/pull/193/commits/a945906066e35af077cbdcbf0cfdccb55ea2eeb3#diff-bd05072ed823b6605df5a6f4a81cdaa5985dd3eec497920bb8ce31cb9abaa7c4R123-R127.

tests/kokkos-based/helpers.hpp
mhoemmen: > Would you mind if I use the reducer's `join()` directly ?

I would prefer that, as a way to avoid code duplication -- thanks!

tests/kokkos-based/helpers.hpp
```diff
@@ -108,52 +108,137 @@ mdspan_t make_mdspan(ValueType *data, std::size_t ext0, std::size_t ext1) {
...
+RealType abs2rel_diff(RealType abs_diff, RealType norm1, RealType norm2)
+{
+  constexpr auto zero = static_cast<RealType>(0);
+  if (norm1 != zero and norm2 != zero) {
+    return abs_diff / std::min(norm1, norm2); // pick larger relative error
+  } else if (norm1 == zero and norm2 == zero) {
+    return zero; // no difference
+  }
+  // Can't get good relative diff with zero -
+  // so return absolute diff out of better ideas...
```
mhoemmen, line 165: This is fine.

tests/kokkos-based/helpers.hpp
```diff
@@ -108,52 +108,137 @@ mdspan_t make_mdspan(ValueType *data, std::size_t ext0, std::size_t ext1) {
...
   if (size != v2.extent(0))
     return false;
   const auto v1_view = KokkosKernelsSTD::Impl::mdspan_to_view(v1);
   const auto v2_view = KokkosKernelsSTD::Impl::mdspan_to_view(v2);
-  // Note: reducint to `int` because Kokkos can complain on `bool` not being 
+  // Note: reducing to `int` because Kokkos can complain on `bool` not being
   //       aligned with int32 and deny it for parallel_reduce()
   using diff_type = int;
   diff_type is_different = false;
+  const auto red = Kokkos::LOr<diff_type>(is_different);
```
mhoemmen, line 214: It's fine to create the reducer in line, but this is fine too : - ) .

mhoemmen (approved):

tests/kokkos-based/helpers.hpp
```diff
@@ -108,8 +108,37 @@ mdspan_t make_mdspan(ValueType *data, std::size_t ext0, std::size_t ext1) {
...
+KOKKOS_INLINE_FUNCTION
+RealValue scalar_abs_diff(RealValue v1, RealValue v2) {
+  return std::abs(v2 - v1);
+}
+
+template <typename T>
+T scalar_abs_diff(const std::complex<T> &v1, const std::complex<T> &v2) {
+  const auto dr = scalar_abs_diff(v1.real(), v2.real());
+  const auto di = scalar_abs_diff(v1.imag(), v2.imag());
+  return std::max(dr, di);
```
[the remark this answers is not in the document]
mhoemmen: @dalg24 This is the error measure that LAPACK likes to use for complex numbers.  I think the original motivation was to avoid square roots.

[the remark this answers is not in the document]
dalg24: no it would be sqrt(8)
https://godbolt.org/z/PoqPobhWT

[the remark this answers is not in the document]
MikolajZuzek: @dalg24 Sorry, somehow I didn't catch that you mean the whole routine...

@dalg24 @mhoemmen 
While `sqrt` stays high on my list, my issue was that `Kokkos::abs()` has no overload for `std::complex` and `std::abs()` has no overload for `Kokkos::complex`. You made me realize we wouldn't need `scalar_diff_abs()` if we had `abs()` that handles both complex types (we extended `conj_if_needed()` this way). Should we introduce that ? If so, should it be in a separate header so the Kokkos algorithm implementations can also use it ?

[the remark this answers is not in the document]
MikolajZuzek: Ok, we don't need extra `abs()` for this: please see https://github.com/kokkos/stdBLAS/pull/193/commits/80a8e2cf24bab00be602c99c6c1cc43259d01f2d. Thank you @dalg24!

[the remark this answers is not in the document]
mhoemmen: @MikolajZuzek It's OK to work around the missing Kokkos feature for now, but any changes to `Kokkos::abs` should happen in Kokkos proper, not here.  Otherwise, you'll get build errors here if `Kokkos::abs` ever adds an overload for `std::complex`.

In brief: Please do not add overloads to Kokkos functions (like `Kokkos::abs`) here.  The Kokkos repository is the right place to add them.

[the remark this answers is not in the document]
MikolajZuzek: Thanks @mhoemmen - totally agree. Fortunately, on https://github.com/kokkos/stdBLAS/commit/80a8e2cf24bab00be602c99c6c1cc43259d01f2d, I was able to just use `std::abs()` and `Kokkos::abs()` instead of introducing new thing for scalar diff - which is what, I think, @dalg24 originally suggested.

tests/kokkos-based/helpers.hpp
```diff
@@ -166,60 +230,249 @@ bool is_same_vector(
...
+          typename AccessorPolicy1,
+          typename ElementType2,
+          std::size_t Extent2,
+          typename LayoutPolicy2,
+          typename AccessorPolicy2>
+auto vector_abs_diff(
+    mdspan<ElementType1, extents<Extent1>, LayoutPolicy1, AccessorPolicy1> v1,
+    mdspan<ElementType2, extents<Extent2>, LayoutPolicy2, AccessorPolicy2> v2)
+{
+  using RetType = decltype(std::abs(v1[0] - v2[0])); // will be same for views
```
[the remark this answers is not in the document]
MikolajZuzek: Unfortunately, that fails with:
```shell
tests/kokkos-based/helpers.hpp:245:39: error: no matching function for call to ‘abs(std::complex<double>)’
  245 |   using RetType = decltype(Kokkos::abs(v1[0] - v2[0])); // will be same for views
      |                            ~~~~~~~~~~~^~~~~~~~~~~~~~~
```
The trick is to call `Kokkos::abs()` for views and `std::abs()` for `mdpan`s. This is exactly why I was tempted to implement `Impl::abs()` that works for both (and dispatches to STL and Kokkos underneath) and that versatility is what `Impl::scalar_abs_diff()` provided.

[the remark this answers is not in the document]
MikolajZuzek: Fixed on https://github.com/kokkos/stdBLAS/pull/193/commits/ea59884bf530f51f94a834973b7cfd6e2471448c - thanks @mhoemmen for pointing this out!

tests/kokkos-based/helpers.hpp
```diff
@@ -166,60 +230,249 @@ bool is_same_vector(
...
+          typename ElementType2,
+          std::size_t Extent20,
+          std::size_t Extent21,
+          typename LayoutPolicy2,
+          typename AccessorPolicy2>
+auto matrix_abs_diff(
+    mdspan<ElementType1, extents<Extent10, Extent11>, LayoutPolicy1, AccessorPolicy1> A,
+    mdspan<ElementType2, extents<Extent20, Extent21>, LayoutPolicy2, AccessorPolicy2> B)
+{
+  using RetType = decltype(std::abs(A(0, 0) - B(0, 0))); // will be same for views
```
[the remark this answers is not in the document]
MikolajZuzek: See [above](https://github.com/kokkos/stdBLAS/pull/193#discussion_r864174636).

mhoemmen (approved):
Thanks for the fixes!

MikolajZuzek:
@mhoemmen @dalg24 Thank you for thorough reviews and great suggestions!

PR #194, opened 2022-04-30
[WIP] kokkos impl `triangular_matrix_vector_solve`

MikolajZuzek:
**WIP:** ready for review (see https://github.com/kokkos/stdBLAS/commit/05f29658b83359b2486d8dc3766451915a5c7fb3) but based on #193 (for relative error in test)

----

Kokkos implementation and unit test for `triangular_matrix_vector_solve()` (not-in-place) based on Kokkos-Kernels [`TeamTRSV`](https://github.com/kokkos/kokkos-kernels/wiki/List-Functor-level-Interface-in-Batched-Functions#Trsv) in [batched algorithms](https://github.com/kokkos/kokkos-kernels/wiki/Batched-BLAS-and-LAPACK).

> **Note:** we don't have [IN-PLACE overload](www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p1673r7.html#in-place-triangular-solve-linalgalgsblas2trsvin-place) - might be related to #134/#136/#139


tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_triangular_matrix_vector_solve.hpp
```diff
@@ -0,0 +1,125 @@
...
+  // convert mdspans to views
+  const auto A_view = Impl::mdspan_to_view(A);
+  const auto b_view = Impl::mdspan_to_view(b);
+  auto x_view = Impl::mdspan_to_view(x);
+
+  // using in-place routine on x=b
+  Kokkos::deep_copy(x_view, b_view);
+
+  // promote x to Nx1 multivector to use KokkosBlas::trsm()
+  // (trsv is not implemented in KokkosBlas)
```
mhoemmen, line 113: Would you consider actually implementing `KokkosBlas::trsv`?  It seems like you're already doing that here.

mhoemmen (approved):

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_triangular_matrix_vector_solve.hpp
fnrizzi: I agree !

tests/kokkos-based/triangular_matrix_vector_solve.cpp
```diff
@@ -0,0 +1,122 @@
...
+namespace{
+
+using namespace kokkostesting;
+
+template<class A_t,
+         class b_t,
+         class x_t,
+         class Triangle,
+         class DiagonalStorage>
+void triangular_matrix_vector_solve_gold_solution(A_t A, Triangle t, DiagonalStorage d, b_t b, x_t x)
```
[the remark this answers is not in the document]
fnrizzi: I am strongly in favor of "gold" self-contained solutions because when we move all kokkos impl to kernels, we will need them anyway for tests

PR #195, opened 2022-04-30
[ready] Refactor shared utilities in Kokkos implementation

MikolajZuzek:
Inspired by #179 for #173, let's consider refactoring of utilities shared by Kokkos algorithm implementations into respective separate headers:
* _triangle.hpp_: various checks and utilities related to triangular matrix handling, including KK tag mappers;
* _static_extent_match.hpp_: for `ParallelMatrixVisitor` (2D `parallel_for` iterating over whole matrix or selected triangle);
* _parallel_matrix.hpp_: `static_extent_match()`;

### DIFF
--- a/tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_matrix_rank_2_update.hpp
+++ b/tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_matrix_rank_2_update.hpp
@@ -47,6 +47,7 @@
 #include <complex>
 #include "signal_kokkos_impl_called.hpp"
 #include "static_extent_match.hpp"
+#include "triangle.hpp"
 
 namespace KokkosKernelsSTD {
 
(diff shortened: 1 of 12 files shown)


tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_matrix_rank_2_update.hpp
```diff
@@ -210,7 +106,7 @@ void symmetric_matrix_rank_2_update(kokkos_exec<ExecSpace> &&exec,
   const auto x_view = Impl::mdspan_to_view(x);
   const auto y_view = Impl::mdspan_to_view(y);
   auto A_view = Impl::mdspan_to_view(A);
-  mtxr2update_impl::ParallelMatrixVisitor v(ExecSpace(), A_view);
+  Impl::ParallelMatrixVisitor v(ExecSpace(), A_view);
```
mhoemmen, line 109: It will be troublesome in the future to have to replace all these `ExecSpace()` with the actual execution policy.  Would you consider changing `ParallelMatrixVisitor` so that it takes the execution policy directly, rather than `ExecSpace`?  That would reduce the number of places where you would have to change `ExecSpace()` to something that gets the wrapped CUDA stream out.

mhoemmen (approved):
Thanks!

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas2_matrix_rank_2_update.hpp
mhoemmen: This can be saved for a future PR, btw.  Thanks!

PR #200, opened 2022-05-04
[ready] kokkos impl: `{symmetrix,hermitian}_matrix_rank_2k_update`

MikolajZuzek:
This PR introduces Kokkos implementation and unit test for [`symmetric_matrix_rank_2k_update()`](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p1673r7.html#rank-2k-symmetric-matrix-update-linalgalgblas3rank2ksyr2k) and [`hermitian_matrix_rank_2k_update()`](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p1673r7.html#rank-2k-hermitian-matrix-update-linalgalgblas3rank2kher2k).

### DIFF
--- a/tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas3_matrix_rank_2k_update.hpp
+++ b/tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas3_matrix_rank_2k_update.hpp
@@ -1,0 +1,162 @@
+
+#ifndef LINALG_TPLIMPLEMENTATIONS_INCLUDE_EXPERIMENTAL___P1673_BITS_KOKKOSKERNELS_MATRIX_RANK2K_UPDATE_HPP_
+#define LINALG_TPLIMPLEMENTATIONS_INCLUDE_EXPERIMENTAL___P1673_BITS_KOKKOSKERNELS_MATRIX_RANK2K_UPDATE_HPP_
+
+/* #include <complex>
+#include "signal_kokkos_impl_called.hpp"
+#include "static_extent_match.hpp"
+#include "triangle.hpp"
+#include "parallel_matrix.hpp" */
+
+namespace KokkosKernelsSTD {
+
+// Rank-2k update of a symmetric matrix
+// performs BLAS xSYR2K: C += A*trans(B) + B*trans(A)
+
+MDSPAN_TEMPLATE_REQUIRES(class ExecSpace,
+    class ElementType_A,
+    std::experimental::extents<>::size_type numRows_A,
+    std::experimental::extents<>::size_type numCols_A,
+    class Layout_A,
+    class ElementType_B,
+    std::experimental::extents<>::size_type numRows_B,
+    std::experimental::extents<>::size_type numCols_B,
+    class Layout_B,
+    class ElementType_C,
+    std::experimental::extents<>::size_type numRows_C,
+    std::experimental::extents<>::size_type numCols_C,
+    class Layout_C,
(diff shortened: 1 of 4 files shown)


MikolajZuzek:
@mhoemmen @fnrizzi 

Does [`symmetric_matrix_rank_2k_update()`](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p1673r7.html#rank-2k-symmetric-matrix-update-linalgalgblas3rank2ksyr2k) allow `A.extent(1) != C.extent(0)` or `B.extent(0) != C.extent(0)` ?

* If yes: shall we fix [the serial implementation](https://github.com/kokkos/stdBLAS/blob/main/include/experimental/__p1673_bits/blas3_matrix_rank_2k_update.hpp#L122-L130) that would crash on `A.extent(1) > B.extent(1)` with `k` going out of bounds in `B(i,k)` ?
* If not: are the _Preconditions_ - which don't mention neither of `A.extent(1)` or `B.extent(0)` - underconstrained ? And what would be the reason for the description to distinguish between `k1` and `k2`, with their respective domains ?

mhoemmen:
@MikolajZuzek You're absolutely right -- I totally forgot to include that in the Preconditions (for dynamic extents) and Mandates (for static extents).  I'll file an issue.  [`DSYR2K` documents](http://www.netlib.org/lapack/explore-html/d1/d54/group__double__blas__level3_ga51f6c38a43599163e670f0f6f5645587.html) that A and B need to have the same number of columns K.

MikolajZuzek:
Thanks @mhoemmen for the clarification!

If `C` is NxN and `A`,`B` are NxK then incorrect _Precondition_ `B.extent(1) equals C.extent(0)` should be replaced with `B.extent(0) equals C.extent(0)` and new one `A.extent(1) equals B.extent(1)` should be added, right ?

PS. Seems like [the hermitian variant](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p1673r7.html#rank-2k-hermitian-matrix-update-linalgalgblas3rank2kher2k) suffers from the same issue, so I took the liberty to update #201 title.

mhoemmen:
@MikolajZuzek Please contact me offline if you would like to be credited in the proposal for finding this issue, and if so, whether you would like to supply an affiliation and/or e-mail address.  Thanks!

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas3_matrix_rank_2k_update.hpp
```diff
@@ -0,0 +1,162 @@
...
+  Impl::signal_kokkos_impl_called("symmetric_matrix_rank_2k_update");
+
+  // convert mdspans to views
+  const auto A_view = Impl::mdspan_to_view(A);
+  const auto B_view = Impl::mdspan_to_view(B);
+  auto C_view = Impl::mdspan_to_view(C);
+
+  using size_type = std::experimental::extents<>::size_type;
+  const auto A_ext1 = A.extent(1); // = B.extent(1)
+  Impl::ParallelMatrixVisitor v(ExecSpace(), C_view);
```
mhoemmen, line 76: [optional] If `ParallelMatrixVisitor`'s `for_each_triangle_matrix_element` method is only ever called once per `ParallelMatrixVisitor` object, then would you consider making it a nonmember function and not exposing the construction of an intermediate object?

mhoemmen (approved):

PR #206, opened 2022-05-17
[ready] kokkos impl: `{symmetric,hermitian}_matrix_{left,right}_product()`

MikolajZuzek:
This PR introduces Kokkos implementation and unit tests for `{symmetric,hermitian}_matrix_{left,right}_product()`.

### DIFF
--- a/tests/kokkos-based/helpers.hpp
+++ b/tests/kokkos-based/helpers.hpp
@@ -108,6 +108,17 @@ mdspan_t make_mdspan(ValueType *data, std::size_t ext0, std::size_t ext1) {
   return mdspan_t(data, ext0, ext1);
 }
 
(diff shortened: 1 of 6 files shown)


tests/kokkos-based/helpers.hpp
```diff
@@ -108,6 +108,17 @@ mdspan_t make_mdspan(ValueType *data, std::size_t ext0, std::size_t ext1) {
   return mdspan_t(data, ext0, ext1);
 }
 
+template<class A_t, class ValueType = typename A_t::value_type>
+void set(A_t A, ValueType value)
```
mhoemmen, line 112: (optional) This seems like a helpful utility.  Would you consider overloading it for `mdspan` not of rank 2?

mhoemmen (approved):

PR #208, opened 2022-05-19
fix matrix inf norm that was wrongly changed before

fnrizzi:


tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_matrix_inf_norm_kk.hpp
```diff
@@ -32,13 +32,15 @@ Scalar matrix_inf_norm(kokkos_exec<ExeSpace> /*kexe*/,
 
   Scalar result = {};
   Kokkos::Max<Scalar> reducer(result);
-  Kokkos::parallel_reduce(Kokkos::RangePolicy(ExeSpace(), 0, A_view.extent(0)*A_view.extent(1)),
+  Kokkos::parallel_reduce(Kokkos::RangePolicy(ExeSpace(), 0, A_view.extent(0)),
```
mhoemmen, line 35: (optional) Thanks for fixing this!  Would you consider just using `Kokkos::MDRange` here?

mhoemmen (approved):

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/blas1_matrix_inf_norm_kk.hpp
fnrizzi: for the inf norm, I need to be able to reduce over rows. So I don't think MDRange will work for this. 
But I can use a team policy so that each team handles a row. what do you think of that?

fnrizzi: what I can do is to use a team policy, one row or so per team, and have each team do a reduction over those rows. 
I was actually planning to do these kind of performance improvements later one but first wanted to focus on having a draft impl of everything. that is why i have simple impl in most places.

PR #215, opened 2022-05-24
cmake fix: require Kokkos

MikolajZuzek:
Replace CMake complaints on missing Kokkos _targets_ with clear messages and hints about missing Kokkos _libraries_.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,8 +99,8 @@ add_library(std::linalg ALIAS linalg)
 
 target_link_libraries(linalg INTERFACE std::mdspan)
 if(LINALG_ENABLE_KOKKOS)
-  find_package(Kokkos)
-  find_package(KokkosKernels)
+  find_package(Kokkos REQUIRED)
+  find_package(KokkosKernels REQUIRED)
   target_link_libraries(linalg INTERFACE Kokkos::kokkos)
   target_link_libraries(linalg INTERFACE Kokkos::kokkoskernels)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -99,8 +99,8 @@ add_library(std::linalg ALIAS linalg)
 
 target_link_libraries(linalg INTERFACE std::mdspan)
 if(LINALG_ENABLE_KOKKOS)
-  find_package(Kokkos)
-  find_package(KokkosKernels)
+  find_package(Kokkos REQUIRED)
```
mhoemmen, line 102: This is an excellent change for improving the error messages.  Should we perhaps consider inverting the dependency between `LINALG_ENABLE_KOKKOS` and finding the Kokkos and KokkosKernels packages, though?  That is, should we consider _first_ doing the `find_package` calls (not required), and only then setting `LINALG_ENABLE_KOKKOS`'s default accordingly?

mhoemmen (approved):

CMakeLists.txt
MikolajZuzek: Thanks!
Now the default for `LINALG_ENABLE_KOKKOS` is based on the library detection (like for `BLAS`), so we can have:
* `OFF`: Kokkos disabled;
* `ON`: connect Kokkos and error out if the library is not found;
* not set: search for Kokkos installation and connect it, if found;

fnrizzi (approved):

crtrott (approved):

PR #217, opened 2022-05-24
kokkos: fix `Kokkos::abs()` namespacing

MikolajZuzek:
Fix compilation error on `Kokkos::abs()` not being recognized when stdBLAS is built aginast Kokkos `3.6.00`, because the function is still in `Kokkos::Experimental` there.

### DIFF
--- a/tests/kokkos-based/helpers.hpp
+++ b/tests/kokkos-based/helpers.hpp
@@ -47,7 +47,7 @@
 #include <experimental/mdspan>
 #include <random>
 
-#if KOKKOS_VERSION <= 30600
+#if KOKKOS_VERSION < 30699
 namespace Kokkos {
   using Experimental::abs;
 }
--- a/tests/kokkos-based/helpers.hpp
+++ b/tests/kokkos-based/helpers.hpp
@@ -47,6 +47,12 @@
 #include <experimental/mdspan>
 #include <random>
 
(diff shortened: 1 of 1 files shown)


tests/kokkos-based/helpers.hpp
```diff
@@ -47,6 +47,12 @@
 #include <experimental/mdspan>
 #include <random>
 
+#if KOKKOS_VERSION <= 30600
```
dalg24, line 50: This would fail with the upcoming patch release 3.6.1

MikolajZuzek: Thanks! Should the number be `30700` then or can that change before 3.7 ?

MikolajZuzek: Is there a smart way to solve it ?
For instance can one include whole `Experimental` into `Kokkos` namespace regardless of version ?
(we'd be doing that for tests only here)

fnrizzi: I think we are probably better off waiting for 3.6.1 and in the meantime we just use Kokkos develop

dalg24: I'd go for
```suggestion
#if KOKKOS_VERSION < 30699
```

mhoemmen (approved):

PR #213, opened 2022-05-21
[wip] add kokkos tests to CI

MikolajZuzek:

PR #219, opened 2022-05-26
kokkos: fix build failng with KOKKOS_DEFAULT=OFF

MikolajZuzek:
Fix build failing on missing `kokkos_exec` when `LINALG_ENABLE_KOKKOS` is `ON` but   `LINALG_ENABLE_KOKKOS_DEFAULT` is `OFF`

### DIFF
--- a/include/experimental/__p1673_bits/linalg_execpolicy_mapper.hpp
+++ b/include/experimental/__p1673_bits/linalg_execpolicy_mapper.hpp
@@ -24,7 +24,7 @@ template<class T> inline constexpr bool is_inline_exec_v = is_inline_exec<T>::va
 }
 
 
-#if defined(LINALG_ENABLE_KOKKOS) && defined(LINALG_ENABLE_KOKKOS_DEFAULT)
+#if defined(LINALG_ENABLE_KOKKOS)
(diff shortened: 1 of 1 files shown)


mhoemmen:
Regarding my comment on PR #220: Do you think the failing test could be fixed straightforwardly for now, even if the fix is not the most efficient?  I would rather have slow code for now, than a disabled test.

MikolajZuzek:
@mhoemmen 
>Regarding my comment on PR https://github.com/kokkos/stdBLAS/pull/220: Do you think the failing test could be fixed straightforwardly for now, even if the fix is not the most efficient? I would rather have slow code for now, than a disabled test.

Yes, please check https://github.com/kokkos/stdBLAS/pull/220#issuecomment-1139467276.

This fix shall just prevent `LINALG_ENABLE_KOKKOS_DEFAULT=OFF` builds from failing. On #213 I propose to keep this flag enabled, so that we gain more coverage by also testing Kokkos implementations with native tests too (in Kokkos CI jobs).

crtrott (approved):

crtrott:
I checked that if DEFAULT is OFF it still does not replace std::par etc.

Issue #218, opened 2022-05-25
Infinite recursion on missing implementation


mzuzek:
@mhoemmen @crtrott @fnrizzi 

If an algorithm is called with execution policy that has identical mapping (i.e. `execpolicy_mapper` returns same type as it gets) - like `kokkos_exec<ExecSpace>` - but implementation (or that particular overload) is missing, the dispatcher enters infinite recursion. It results in SEGFAULT crash in `Debug` builds or hangs in `Release` builds (e.g. see [the hanging tests cancelled after 25m](https://github.com/kokkos/stdBLAS/runs/6571220999?check_suite_focus=true)).

This happens because `is_custom_<ALG>_avail` checks mistake _the dispatcher routine_ for requested custom implementation (as - with id policy mapping - their function signatures are identical).

MikolajZuzek:
One possible solution would be to have `is_custom_<ALG>_avail` checks detect that the dispatcher routine is detected as the custom function like:
```cpp
std::is_same_v< std::experimental::linalg::triangular_matrix_left_product, // dispatcher
                    /* ADL namespace :: */ triangular_matrix_left_product >      // detected custom implementation
```
For example:
```cpp
template <class Exec, class A_t, class Tr_t, class DiagSt_t, class C_t>
struct is_custom_triang_mat_left_product_with_update_avail<
  Exec, A_t, Tr_t, DiagSt_t, C_t,
  std::enable_if_t<
    std::is_void_v<decltype(
	       triangular_matrix_left_product
	       (std::declval<Exec>(),
		std::declval<A_t>(),
		std::declval<Tr_t>(),
		std::declval<DiagSt_t>(),
		std::declval<C_t>()))>

    // check if custom implementation is different than the dispatcher to prevent inf recursion
    && !std::is_same_v< 
      decltype(
	       std::experimental::linalg::triangular_matrix_left_product
	       (std::declval<Exec>(),
            std::declval<A_t>(),
            std::declval<Tr_t>(),
            std::declval<DiagSt_t>(),
            std::declval<C_t>())),
      decltype(
	       triangular_matrix_left_product
	       (std::declval<Exec>(),
            std::declval<A_t>(),
            std::declval<Tr_t>(),
            std::declval<DiagSt_t>(),
            std::declval<C_t>()))>

    && !linalg::impl::is_inline_exec_v<Exec>
    >
  >
  : std::true_type{};
```

fnrizzi:
[...]

PR #220, opened 2022-05-26
exclude native `gemm` test in Kokkos+default config

MikolajZuzek:
In build with default Kokkos execution space (both `LINALG_ENABLE_KOKKOS` and `LINALG_ENABLE_KOKKOS_DEFAULT` enabled) the native `gemm` test runs Kokkos implementation and fails on missing overloads for [`transposed`](https://github.com/kokkos/stdBLAS/blob/93a1fa2aead9d2eb723d3587b009c229a623ae98/tests/native/gemm.cpp#L171-L172)- and [`scaled`](https://github.com/kokkos/stdBLAS/blob/93a1fa2aead9d2eb723d3587b009c229a623ae98/tests/native/gemm.cpp#L217-L218)-wrapped inputs - see #218.

One possible workaround is to disable the test in considered scenario, if we don't plan to support the problematic accessors in the Kokkos implementation any soon.

On this PR I'm proposing generic test exclusion list (`STDBLAS_TESTS_EXCLUDE`) to avoid "polluting" native tests CMake scripting with Kokkos code, which would happen with simpler solution like:
```cmake
if(NOT (LINALG_ENABLE_KOKKOS AND LINALG_ENABLE_KOKKOS_DEFAULT))
  linalg_add_test(gemm)
endif()
```

> Detected by tests [to be] failing on [CI](https://github.com/kokkos/stdBLAS/runs/6581885052?check_suite_focus=true): `7 - gemm (SEGFAULT)`


mhoemmen:
> One possible workaround is to disable the test in considered scenario, if we don't plan to support the problematic accessors in the Kokkos implementation any soon.

The `scaled` and `transposed` input matrix modifiers are fundamental to the proposal.  Instead of disabling the failing test, would you consider a temporary fix that falls back to the sequential implementation if it encounters these accessors?

MikolajZuzek:
@mhoemmen 
> Instead of disabling the failing test, would you consider a temporary fix that falls back to the sequential implementation if it encounters these accessors?

Sure, please see https://github.com/kokkos/stdBLAS/pull/222. I've created new PR for the alternative solution, so this one can be just closed to mark this approach as rejected.

mhoemmen:
@MikolajZuzek Thanks for explaining!  I'll close this PR then.

PR #230, opened 2022-06-07
Permit C++ < 20

mhoemmen:
kokkos/stdBLAS is now known to compile for C++ versions less than 2020.  Update CMakeLists.txt so that it respects the user's `LINALG_CXX_STANDARD` setting, rather than forcing it to "20" always.  I've tested that it accepts "17".  This also brings stdBLAS's CMake logic in line with mdspan's.


nliber (approved):

mhoemmen:
Thanks @nliber ! : - D

PR #232, opened 2022-06-07
Fix GCC 9 build by removing TBB requirement

### LINKED ISSUE #231: Build fails, requiring TBB, with GCC < 10
Building stdBLAS with GCC 9.1 currently fails, because linalg_execpolicy_mapper.hpp unconditionally includes `<execution>`.  This, in turn, unconditionally includes a TBB (Intel Threading Building Blocks) header file.

The following godbolt link and code example demonstrates that GCC < 10 breaks when including `<execution>`: https://godbolt.org/z/b3rr5qvYb
```c++
#include <iostream>

#if ! defined(__GNUC__) || __GNUC__ > 9
#include <execution>
#endif

int main()
{
    std::cout << "Hello world!" << std::endl;
    return 0;
}
```
Typical error message:
```
In file included from /opt/compiler-explorer/gcc-9.4.0/include/c++/9.4.0/pstl/parallel_backend.h:14,
                 from /opt/compiler-explorer/gcc-9.4.0/include/c++/9.4.0/pstl/algorithm_impl.h:25,
                 from /opt/compiler-explorer/gcc-9.4.0/include/c++/9.4.0/pstl/glue_execution_defs.h:52,
                 from /opt/compiler-explorer/gcc-9.4.0/include/c++/9.4.0/execution:32,
                 from <source>:4:
/opt/compiler-explorer/gcc-9.4.0/include/c++/9.4.0/pstl/parallel_backend_tbb.h:19:10: fatal error: tbb/blocked_range.h: No such file or directory
   19 | #include <tbb/blocked_range.h>
      |          ^~~~~~~~~~~~~~~~~~~~~
```
Suggested fix: Protect inclusion of `<execution>` with `#if defined(__GNUC__) && __GNUC__ > 9 ... #endif`.

mhoemmen:
Fixes #231.


nliber (approved):

mhoemmen:
Thanks @nliber ! : - D

Issue #233, opened 2022-06-08
transposed: Implement R8/9 return type rules


mhoemmen:
`transposed` for many layouts doesn't need to fall back to `layout_transpose`.  It can just flip the extents and layout (e.g., row major to column major, or vice versa).  R8 and R9 of P1673 specify this precisely; we just need to implement it.

mhoemmen:
Fixed by PR #238.

PR #234, opened 2022-06-09
Bring reference implementation in line with P1673R9

mhoemmen:
Bring the reference implementation of P1673 in line with P1673R9 (submitted 2022/06/14).

Most of the changes relate to `conjugated`, `conjugate_transposed`, `scaled`, and `transposed` (and the layout and accessors that implement them).

Fixes #135, #233, and other issues.

Edit (2022/06/19): This is now rebased atop PR #236.  Please merge PR #236 first before merging this PR.  (This PR will be unlikely to build otherwise.)


include/experimental/__p1673_bits/conjugated.hpp
```diff
@@ -44,135 +44,92 @@
...
-  accessor_conjugate() = default;
-
-  accessor_conjugate(Accessor a) : acc(a) {}
-
-  reference access(pointer p, size_type i) const noexcept {
-    return reference(acc.access(p, i));
+    std::conditional_t<accessor_value_is_arith,
+		       typename Accessor::offset_policy,
+		       accessor_conjugate<typename Accessor::offset_policy>>;
+
```
youyu3, line 82: It would be good to keep the default constructor - without it something in my code is broken.

include/experimental/__p1673_bits/scaled.hpp
```diff
@@ -50,87 +50,78 @@ namespace experimental {
...
-
-  accessor_scaled() = default;
-
-  accessor_scaled(Accessor a, S sval) :
-    acc_(a), scale_factor_(sval) {}
-
+  using offset_policy =
+    accessor_scaled<ScalingFactor, typename Accessor::offset_policy>;
+
+  accessor_scaled(ScalingFactor scaling_factor, Accessor accessor) :
```
youyu3, line 64: It would be good to keep the default constructor - without it something in my code is broken.

include/experimental/__p1673_bits/conjugated.hpp
```diff
@@ -44,135 +44,92 @@
...
+      typename Accessor::element_type,
+      typename reference::value_type>;
+  
 public:
-  // FIXME If BLAS functions want to strip off accessor_conjugate for
-  // optimization, they will need a way to work with th underlying
-  // Accessor (which may not be the default one).
-
-  using element_type  = typename Accessor::element_type;
+  using element_type  = std::add_const_t<pre_const_element_type>;
```
youyu3, line 76: So the idea here is that you don't allow people to modify the values through something like `auto conj_a = conjugated(a); conj_a(0) = 0;`?

tests/native/scaled.cpp
```diff
@@ -1,14 +1,37 @@
...
+    EXPECT_EQ(x_scaled[0], (1 << 20) * d);
+    auto x_scaled_scaled = scaled(1 << 20, x_scaled);
+    EXPECT_EQ(x_scaled_scaled[0], expected_val);
+    static_assert(std::is_same_v<decltype(x_scaled_scaled)::value_type, double>);
+    static_assert(std::is_same_v<decltype(x_scaled_scaled)::element_type, const double>);    
+  }
+
   TEST(scaled, mdspan_double_scalar_float)
   {
     using vector_element_type = double;
```
youyu3, line 37: There are compilation errors from `proxy_reference.hpp` when both `vector_element_type` and `scaling_factor_type` are of `std::complex<double>`.  I suppose this should work?

```
/home/yu/repos/nv/stdblas_test/stdBLAS/include/experimental/__p1673_bits/proxy_reference.hpp:163:46: error: no match for ‘operator+’ (operand types are ‘std::experimental::__p1673_version_0::linalg::impl::proxy_reference<std::complex<double>&, std::complex<double>, std::experimental::__p1673_version_0::linalg::scaled_scalar<std::complex<double>, std::complex<double>&, std::complex<double> > >::value_type’ {aka ‘std::complex<double>’} and ‘float’)
  163 |   P1673_PROXY_REFERENCE_ARITHMETIC_OPERATOR( + )
      |                                              ^
```

tests/native/scaled.cpp
```diff
@@ -1,14 +1,37 @@
+#include "gtest/gtest.h"
```
youyu3, line 1: I also tried adding a test case for `std:complex`
```
  TEST(scaled, mdspan_complex)
  {
    using vector_element_type = std::complex<double>;
    using scaling_factor_type = std::complex<double>;

    using vector_t =
    mdspan<vector_element_type, extents<dynamic_extent>>;

    constexpr std::size_t vectorSize (5);
    constexpr std::size_t storageSize = vectorSize;
    std::vector<vector_element_type> storage (storageSize);

    vector_t x (storage.data (), vectorSize);

    for (std::size_t k = 0; k < vectorSize; ++k) {
      const vector_element_type x_k = vector_element_type(k) + 1.0;
      x(k) = x_k;
    }   

    const scaling_factor_type scalingFactor (-3.0);
    auto x_scaled = scaled (scalingFactor, x); 

    using std::abs;

    for (std::size_t k = 0; k < vectorSize; ++k) {
      auto abs_x_scaled_k = abs(x_scaled(k));
    }   
  }
```
and it ran into the compilation error that I had before (I needed to add a `static_cast` get this to work)
```
/home/yu/repos/nv/stdblas_test/stdBLAS/tests/native/scaled.cpp:42:44: error: no matching function for call to ‘abs(std::experimental::mdspan<const std::complex<double>, std::experimental::extents<18446744073709551615>, std::experimental::layout_right, std::experimental::__p1673_version_0::linalg::accessor_scaled<std::complex<double>, std::experimental::default_accessor<std::complex<double> > > >::reference)’
   42 |       auto abs_x_scaled_k = abs(x_scaled(k));
      |                                            ^
```

include/experimental/__p1673_bits/conjugated.hpp
mhoemmen: @youyu3 Default constructibility isn't one of the accessor policy requirements, so I would prefer not to require it here.  Please coordinate with me offline to see if it's possible to work around.

include/experimental/__p1673_bits/scaled.hpp
mhoemmen: Please see note above; thanks!

include/experimental/__p1673_bits/conjugated.hpp
mhoemmen: @youyu3 Yes, absolutely.  `conjugated` was already specified to return a read-only view.  I realized when going through the implementation that I could just `add_const_t` the element type.  I didn't need the "nonconst version of the nested accessor," because I'm not changing the nested accessor; I'm returning a new accessor whose element type is the const version of the nested accessor's element type.

tests/native/scaled.cpp
youyu3, line 1: Also there seems to be a problem when the element value and scaling factor are both of `int8_t`.  The element type of the scaled mdspan seems to be `int` not `int8_t`.
```
  TEST(scaled, mdspan_int8)
  {
    using vector_element_type = std::int8_t;
    using scaling_factor_type = std::int8_t;

    using vector_t =
    mdspan<vector_element_type, extents<dynamic_extent>>;

    constexpr std::size_t vectorSize (5);
    constexpr std::size_t storageSize = vectorSize;
    std::vector<vector_element_type> storage (storageSize);

    vector_t x (storage.data (), vectorSize);

    for (std::size_t k = 0; k < vectorSize; ++k) {
      const vector_element_type x_k = vector_element_type(k) + 1.0;
      x(k) = x_k;
    }

    const scaling_factor_type scalingFactor (-3.0);
    auto x_scaled = scaled (scalingFactor, x);

    using val_t = typename decltype(x_scaled)::value_type;
    //static_assert( std::is_same_v< std::remove_cv_t<val_t>, int > );
    static_assert( std::is_same_v< std::remove_cv_t<val_t>, vector_element_type > );
  }
```

tests/native/scaled.cpp
mhoemmen: @youyu3 Could that `float` perhaps come from `scaled` in a test?  The only thing that could be mixing `float` and `complex<double>` would be a test.

tests/native/scaled.cpp
mhoemmen: @youyu3 Let me take a look at this one.  Thanks for reviewing and for the test case!

mhoemmen: @youyu3 The implementation is following the usual integer promotion rules.

https://godbolt.org/z/Ec3zeGqP9
```c++
#include <type_traits>
#include <cstdint>

int main()
{
    int8_t scalingFactor = 3.0; // narrowed
    int8_t x = 4;
    auto d_times_i = scalingFactor * x; // what scaled_scalar does
    static_assert(std::is_same_v<decltype(d_times_i), int>);  
    return 0;
}
```

youyu3: Ah, good to know that integer promotion rules are designed in that way.  This makes sense, but I'll need to fix my code.

tests/native/scaled.cpp
youyu3: Oh didn't read what the test was doing.  Yes you are right :).

tests/native/scaled.cpp
mhoemmen: @youyu3 In general, I won't be able to make `scaled_scalar` a replacement for its underlying `value_type` in every possible expression.  I can add a hidden friend `abs` that makes this particular expression work for unqualified `abs`.  That seems reasonable, given that the result of `scaled` could be the input of `vector_norm_2`.  What do you think?

youyu3: Yeah that sounds reasonable to me if there is no better solution.  `scaled` could be the input of a few other blas1 functions as well.  Also we might need this treatment for other functions that can be called with both non-complex and complex data types, like `exp` and `pow`, although they are not used in  stdblas functions themselves?

mhoemmen: Yup, it's a little weird that the product of two `int8_t` is `int`, but such is the weird language we find ourselves using ; - )

mhoemmen: @youyu3 wrote:
> Also we might need this treatment for other functions that can be called with both non-complex and complex data types, like exp and pow, although they are not used in stdblas functions themselves?

The result of `scaled`, `conjugated`, `transposed`, or `conjugate_transposed` only needs to be usable inside stdblas functions.  It's an interesting side effect that it happens to behave in a certain way outside stdblas functions, but that was never the goal of these functions.  We do need `abs` to work generally because of `vector_abs_norm`.

mhoemmen: @youyu3 I've added `abs`, `real`, and `imag` to `proxy_reference`.  This should make `vector_abs_sum` work.

include/experimental/__p1673_bits/conjugated.hpp
mhoemmen: Update: it was possible to work around, yay!

include/experimental/__p1673_bits/scaled.hpp
mhoemmen: Update: it was possible to work around without changing the spec or reference implementation, yay!

crtrott (changes requested):

include/experimental/__p1673_bits/conjugated.hpp
```diff
@@ -44,135 +44,92 @@
...
+  accessor_conjugate(Accessor accessor) : accessor_(accessor) {}
+
+  MDSPAN_TEMPLATE_REQUIRES(
+    class OtherElementType,
+    /* requires */ (std::is_convertible_v<
+      typename default_accessor<OtherElementType>::element_type(*)[],
+      typename Accessor::element_type(*)[]
+    >)
+  )
+  accessor_conjugate(default_accessor<Accessor> accessor) : accessor_(accessor) {}
```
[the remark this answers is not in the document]
mhoemmen: (Repeating offline discussion) Your observation is correct -- `Accessor` should not go inside `default_accessor`; this code is incorrect.  The intent is conversion from nonconst element type to const element type, for the special case of `default_accessor` (since there is currently no generic way to get the const-element version of a nonconst-element accessor).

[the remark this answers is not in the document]
mhoemmen: @crtrott I fixed this and added a test.  Thanks for the catch!

mhoemmen:
@crtrott The `accessor_conjugated` issue you pointed out is fixed.  This PR is now rebased atop PR https://github.com/kokkos/stdBLAS/pull/236. Please merge PR https://github.com/kokkos/stdBLAS/pull/236 first before merging this PR. (This PR will be unlikely to build otherwise.)

PR #236, opened 2022-06-17
Use mdspan-0.3.0 tag

mhoemmen:
mdspan will soon experience a backwards compatibility - breaking change.
Please see info here: https://github.com/kokkos/mdspan/issues/136
This commit temporarily changes stdBLAS to use the indicated pre-change mdspan tag for github-based tests,
and to fetch that tag (rather than the stable tag) when CMake needs to fetch and build mdspan.

### DIFF
--- a/examples/01_scale.cpp
+++ b/examples/01_scale.cpp
@@ -2,7 +2,7 @@
 
 #include <iostream>
 
-#if ! defined(__GNUC__) || __GNUC__ > 9
+#if (! defined(__GNUC__)) || (__GNUC__ > 9)
 #  define MDSPAN_EXAMPLES_USE_EXECUTION_POLICIES 1
 #endif
 
--- a/examples/01_scale.cpp
+++ b/examples/01_scale.cpp
@@ -2,6 +2,14 @@
 
 #include <iostream>
 
+#if ! defined(__GNUC__) || __GNUC__ > 9
+#  define MDSPAN_EXAMPLES_USE_EXECUTION_POLICIES 1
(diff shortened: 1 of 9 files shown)


examples/02_matrix_vector_product_basic.cpp
```diff
@@ -2,6 +2,14 @@
 
 #include <iostream>
 
+#if ! defined(__GNUC__) || __GNUC__ > 9
```
crtrott, line 5: is this correct?

crtrott: I.e. do you need parenthesis?

examples/01_scale.cpp
```diff
@@ -2,6 +2,14 @@
 
 #include <iostream>
 
+#if ! defined(__GNUC__) || __GNUC__ > 9
```
crtrott, line 5: parenthesis needed?

crtrott (changes requested):
Do we need parenthesis?

examples/02_matrix_vector_product_basic.cpp
mhoemmen: I don't think so, but I'll put parens in.

mhoemmen: Fixed!

examples/01_scale.cpp
mhoemmen: Fixed!

crtrott (approved):

PR #239, opened 2022-06-21
Remove caching of mdspan

### LINKED ISSUE #237: github action does not use mdspan SHA1 correctly
In PR #234, we had to change the mdspan key (in .github/workflows/cmake.yml) so that the PR testing wouldn't always pick up an old mdspan (with the wrong tag, stable instead of mdspan-0.3.0).  The workflow output suggested that for `key: mdspan-${{ steps.get-sha.outputs.sha }}`, the key was evaluating to `mdspan-null`.  Thus, it could be that the code for getting the SHA1 is broken.  If we don't fix this, then we'll never be able to test stdBLAS against the latest mdspan; it will always hit the cache, unless we change it manually (as we had to do in PR #234).

amklinv-nnl:
Fixes #237

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -16,39 +16,24 @@ jobs:
         - compiler_driver: g++
           compiler_prefix: /usr/bin
     steps:
-    - id: get-sha
-      run: echo ::set-output name=sha::$( curl https://api.github.com/repos/kokkos/mdspan/git/ref/heads/stable | jq .object.sha | tr -d '"' )
- 
-    - name: Determine whether mdspan needs to be rebuilt
-      id: cache-mdspan
-      uses: actions/cache@v2
-      with:
-        path: mdspan-install
-        key: mdspan-stable-${{ steps.get-sha.outputs.sha }}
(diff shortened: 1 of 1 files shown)


amklinv-nnl:
@mhoemmen

.github/workflows/cmake.yml
```diff
@@ -16,39 +16,24 @@ jobs:
...
     - name: Create Build Environment
-      if: steps.cache-mdspan.outputs.cache-hit != 'true'
       run: cmake -E make_directory mdspan-build
       
     - name: Check Out
-      if: steps.cache-mdspan.outputs.cache-hit != 'true'
       uses: actions/checkout@v2
       with:
         repository: kokkos/mdspan
         path: mdspan-src
```
mhoemmen, line 26: Is there a way to ask for a specific branch or tag of mdspan here?  We don't immediately need that feature, but it could be useful in the future.  We could just set it to "stable" (lol) for now.

mhoemmen (approved):
@amklinv-nnl Thanks so much!!! : - D

PR #245, opened 2022-07-12
Fix triangular_matrix_matrix_solve loop & build error

### LINKED ISSUE #244: Bug in `triangular_matrix_matrix_solve`
In `trsm_lower_triangular_right_side`, the loop over index 'j' should have a decreasing `j`, i.e., `for (ptrdiff_t j =  A_num_cols - 1; j >= 0; --j)`.

### LINKED ISSUE #246: Question regarding conj_if_needed
At line 87 in conjugated_if_needed.hpp, `conj_if_needed` is implemented as:
```
auto conj_if_needed = [](const auto& t)
{
  using T = std::remove_const_t<decltype(t)>;
  return conj_if_needed_impl(t, has_conj<T>{});
};
```
Is this code missing `template <class T>` or `inline`? As the code is now, it will be defined multiple times.

mhoemmen:
@youyu3 

The sequential implementation of `triangular_matrix_matrix_solve` for the lower triangular, right side case had a j loop that was counting in the wrong direction: up, instead of down.  This commit fixes that, and adds a regression test.  The test also uncovered a compilation bug that this commit fixes.

I also took the opportunity to explore a new design for avoiding duplicated code in defining template parameters and their corresponding function parameters.

In addition, I added a commit that fixes #246.

Fixes #244 and #246.


include/experimental/__p1673_bits/blas3_triangular_matrix_matrix_solve.hpp
```diff
@@ -188,35 +167,29 @@ void trsm_upper_triangular_right_side(
...
   DiagonalStorage d,
-  std::experimental::mdspan<ElementType_B, std::experimental::extents<SizeType_B, numRows_B, numCols_B>, Layout_B, Accessor_B> B,
-  std::experimental::mdspan<ElementType_X, std::experimental::extents<SizeType_X, numRows_X, numCols_X>, Layout_X, Accessor_X> X)
+  P1673_MATRIX_PARAMETER( B ),
+  P1673_MATRIX_PARAMETER( X ))
 {
   constexpr bool explicit_diagonal =
     std::is_same_v<DiagonalStorage, explicit_diagonal_t>;
   using size_type = ::std::common_type_t<SizeType_A, SizeType_B, SizeType_X>;
+  using signed_index_type = ::std::make_signed_t<size_type>;
```
youyu3, line 185: Question: is this preferred over `ptrdiff_t` and will/should the usage of `ptrdiff_t` for index decrement at other places be replaced with this?

mhoemmen: @youyu3 Thanks for asking!  The two priorities for the sequential / "fall-back" implementation are

1. avoid unwarranted overflow; and
2. avoid build warnings (e.g., on signed-to-unsigned conversion).

A reasonable implementation approach would be to use `size_t` if the common type of all the index types is unsigned, and `ptrdiff_t` if the common type is signed.  It doesn't matter, as long as we use an index type that can't overflow.

nliber (approved):

Issue #248, opened 2023-03-07
Matrix product always falls back to native implementation 


srinivasyadav18:
I am trying to implement HPX backend for `matrix_product`. 
I am not certainly clear with this section of code (below).
The following always evalutes to false right ?
Because both qualified lookup and unqualified lookup of `matrix_product` return `void`, so the second part of the expression always evaluates to false and hence whole `std::enable_if_t` is false.
```c++
template <class Exec, class A_t, class B_t, class C_t>
struct is_custom_matrix_product_avail<
  Exec, A_t, B_t, C_t,
  std::enable_if_t<
    std::is_void_v<
      decltype(
	      matrix_product(
          std::declval<Exec>(),
          std::declval<A_t>(),
		      std::declval<B_t>(),
		      std::declval<C_t>()))
      >
    && !std::is_same_v< // see #218
      decltype(
        std::experimental::linalg::matrix_product(
          std::declval<Exec>(),
          std::declval<A_t>(),
		      std::declval<B_t>(),
		      std::declval<C_t>())),
      decltype(
        matrix_product(
          std::declval<Exec>(),
          std::declval<A_t>(),
		      std::declval<B_t>(),
		      std::declval<C_t>()))
      >
    && !linalg::impl::is_inline_exec_v<Exec>
    >
  >
  : std::true_type{};
```

Which makes this : https://github.com/kokkos/stdBLAS/blob/main/include/experimental/__p1673_bits/blas3_matrix_product.hpp#L757 always false and hence falls backs to native sequential implementation.

fnrizzi:
hi @srinivasyadav18 , pinging @mzuzek since he is the one who posted #218

mzuzek:
@srinivasyadav18 @fnrizzi 
Hi!
Thanks for rising this.
I've just created #249 which reverts my incorrect fix #222 for #218 - essentially removing the second  condition.
For more details and background, please see the discussion at https://github.com/kokkos/stdBLAS/issues/218#issuecomment-1460039688.

srinivasyadav18:
@fnrizzi @mzuzek 

Thank you.
#249 Works for me now as expected. It is being dispatched to the correct overload now.

PR #249, opened 2023-03-08
Revert PR#222 with incorrect fix for #218

mzuzek:
This PR reverts #222 (commit 098f6ef27e1e79a6634ea1f2ad6bffc91b3f506e) with partial fix for #218:
the implementation was found out to be incorrect - see #248 and discussion on #218 for more details.

### DIFF
--- a/include/experimental/__p1673_bits/blas3_matrix_product.hpp
+++ b/include/experimental/__p1673_bits/blas3_matrix_product.hpp
@@ -368,6 +368,23 @@ constexpr bool extractConj(in_matrix_t A)
 template <class Exec, class A_t, class B_t, class C_t, class = void>
 struct is_custom_matrix_product_avail : std::false_type {};
 
+template <class Exec, class A_t, class B_t, class C_t>
+struct is_custom_matrix_product_avail<
+  Exec, A_t, B_t, C_t,
+  std::enable_if_t<
+    std::is_void_v<
+      decltype(
+	       matrix_product
+	       (std::declval<Exec>(),
+		std::declval<A_t>(),
+		std::declval<B_t>(),
+		std::declval<C_t>()))
+      >
+    && !linalg::impl::is_inline_exec_v<Exec>
+    >
+  >
+  : std::true_type{};
+
 template <class Exec, class A_t, class B_t, class E_t, class C_t, class = void>
 struct is_custom_matrix_product_with_update_avail : std::false_type {};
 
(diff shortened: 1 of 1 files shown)


mhoemmen (approved):
I'll leave this up to @crtrott , who is more familiar with the customization design.  Here's what I wrote on #218.

> I do not like that the reference implementation is trying to ADL-find some customization points with the same names.
> 
> The proposal is still in flight. LEWG asked us to change function names in R12. When we make those changes here, we will break whatever code depends on those names.
>
> The Right Way is to separate stdBLAS's public interface from its implementation. Any customization points should have DIFFERENT NAMES. Users should not have to worry about compiler errors because the call to matrix_vector_product is ambiguous between `std::linalg::matrix_vector_product` and `my_funny_namespace::matrix_vector_product`.

mhoemmen:
@fnrizzi wrote:
> where is your comment in the issue? it disappeared?

I deleted the comment, because it wasn't accurate.  Never mind : - )  Christian and I are discussing the solution now.

crtrott:
Basically the infinite recursion should only happen if the specialization (i.e. the Kokkos variant) isn't more special than the generic one which is completely unconstrained on the ExecPolicy see https://godbolt.org/z/7x5hhne8o

Basically the specialized implementation needs to have the same constraints as the entry function (i.e. you need to take the mdspans as mdspans and be templated on all the other things). In that way this won't recurse.

fnrizzi:
> Basically the infinite recursion should only happen if the specialization (i.e. the Kokkos variant) isn't more special than the generic one which is completely unconstrained on the ExecPolicy see https://godbolt.org/z/7x5hhne8o
> 
> Basically the specialized implementation needs to have the same constraints as the entry function (i.e. you need to take the mdspans as mdspans and be templated on all the other things). In that way this won't recurse.

I was going to say I agree with differentiating names, and in fact I brought it up a while back, but was kind of "not a pressing issue" at that time I guess. why is ADL a better solution?

mhoemmen:
@fnrizzi Just to clarify: Christian and I did some investigation, and we think the actual issue could be that your custom `matrix_vector_product` doesn't sufficiently constrain its parameters.

fnrizzi:
> @fnrizzi Just to clarify: Christian and I did some investigation, and we think the actual issue could be that your custom `matrix_vector_product` doesn't sufficiently constrain its parameters.

yes, ok, i see it could be the case also from what Christian said. But this being said, I don't quite understand why not using a different name? That would take away all issues and it seems to me that relying on ADL and the fact that customizations must be constrained *enough/well* is a kind of a delicate thing, very easy to mess up. I do get that it has to be done once, but still, any small change can inadvertently break this.

mhoemmen:
@fnrizzi wrote:
> But this being said, I don't quite understand why not using a different name? Effectively, this would also express better(IMO) the intent since one is writing an implementation of the stdlinalg and those impl functions should in theory not even be used directly. That would take away all issues and it seems to me that relying on ADL and the fact that customizations must be constrained enough/well is a kind of a delicate thing, ....

Users want to customize algorithms like `matrix_vector_product`, and users also want the stdblas implementation to find customizations if they exist.  However, if users get the customization wrong (e.g., incorrect constraints), then bad things can happen.

Choosing a different name for the customization point than the function would prevent infinite recursion issues like #218.  However, it wouldn't prevent other issues.  For example, if users implementing a customization don't constrain its parameters correctly or provide the wrong number of parameters, then the compiler might silently ignore the customization.

[P2279](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2279r0.html) has a survey of different customization approaches.  It explains the different ways that each of them can go wrong, and argues for a language solution.  [P2547](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2547r1.html) is an example of a proposed language solution.

I wouldn't object to making the customization points have different names.  However, this would be a breaking change, yet it would only solve one of the many issues with relying on ADL as a customization mechanism.  If we plan to make a breaking change, then we might want to consider adopting a different approach for customization.

mhoemmen:
@fnrizzi wrote:
> I was actually reasoning from the assumption that the customization mechanism was "decided" except for the naming...

Just to clarify: the customization mechanism is not part of the proposal P1673.  It's specific to this implementation of the proposal.  Thus, it can be changed independently of the proposal.

PR #251, opened 2023-05-31
Added language standard parallelism to dot product

amklinv-nnl:
@mhoemmen @crtrott Please review at your convenience and I'll do the same for the other functions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,6 +130,15 @@ if(LINALG_ENABLE_KOKKOS)
   find_package(KokkosKernels REQUIRED)
 endif()
 
+find_package(TBB)
+option(LINALG_ENABLE_TBB
+  "Enable Threaded Building Blocks for tests. Default: autodetect TBB installation."
+  ${TBB_FOUND}
+)
+if(LINALG_ENABLE_TBB)
+  find_package(TBB REQUIRED)
+endif()
+
 ################################################################################
 
 CONFIGURE_FILE(include/experimental/__p1673_bits/linalg_config.h.in
@@ -150,6 +159,10 @@ if(LINALG_ENABLE_KOKKOS)
   target_include_directories(linalg INTERFACE
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/tpl-implementations/include>
   )
+endif()
+
+if(LINALG_ENABLE_TBB)
+  target_link_libraries(linalg INTERFACE TBB::tbb)
 endif()
 
 target_include_directories(linalg INTERFACE
--- a/examples/01_scale.cpp
+++ b/examples/01_scale.cpp
@@ -13,7 +13,7 @@
 // Make mdspan less verbose
 using std::experimental::mdspan;
 using std::experimental::extents;
-using std::experimental::dynamic_extent;
+using std::dynamic_extent;
 
 int main(int argc, char* argv[]) {
   std::cout << "Scale" << std::endl;
--- a/examples/02_matrix_vector_product_basic.cpp
+++ b/examples/02_matrix_vector_product_basic.cpp
@@ -13,7 +13,7 @@
 // Make mdspan less verbose
 using std::experimental::mdspan;
 using std::experimental::extents;
-using std::experimental::dynamic_extent;
+using std::dynamic_extent;
 
 int main(int argc, char* argv[]) {
   std::cout << "Matrix Vector Product Basic" << std::endl;
--- a/examples/03_matrix_vector_product_mixedprec.cpp
+++ b/examples/03_matrix_vector_product_mixedprec.cpp
@@ -5,7 +5,7 @@
 // Make mdspan less verbose
 using std::experimental::mdspan;
 using std::experimental::extents;
-using std::experimental::dynamic_extent;
+using std::dynamic_extent;
 using std::experimental::submdspan;
 using std::experimental::full_extent;
 
--- a/include/experimental/__p1673_bits/blas1_dot.hpp
+++ b/include/experimental/__p1673_bits/blas1_dot.hpp
@@ -43,6 +43,7 @@
 #ifndef LINALG_INCLUDE_EXPERIMENTAL___P1673_BITS_BLAS1_DOT_HPP_
 #define LINALG_INCLUDE_EXPERIMENTAL___P1673_BITS_BLAS1_DOT_HPP_
 
+#include <ranges>
 #include <type_traits>
 
 namespace std {
@@ -90,7 +91,7 @@ template<class ElementType1,
          class Accessor2,
          class Scalar>
 Scalar dot(
-  std::experimental::linalg::impl::inline_exec_t&& /* exec */,
+  std::experimental::linalg::impl::inline_exec_t&& exec,
   std::experimental::mdspan<ElementType1, std::experimental::extents<SizeType1, ext1>, Layout1, Accessor1> v1,
   std::experimental::mdspan<ElementType2, std::experimental::extents<SizeType2, ext2>, Layout2, Accessor2> v2,
   Scalar init)
@@ -100,10 +101,18 @@ Scalar dot(
                 v1.static_extent(0) == v2.static_extent(0));
 
   using size_type = std::common_type_t<SizeType1, SizeType2>;
-  for (size_type k = 0; k < v1.extent(0); ++k) {
-    init += v1(k) * v2(k);
-  }
-  return init;
+  using scalar_type = std::common_type_t<ElementType1, ElementType2, Scalar>;
+  using std::ranges::iota_view;
+  using std::ranges::begin;
+  using std::ranges::end;
+
+  iota_view range{size_type{}, v1.extent(0)};
+
+  Scalar sum = std::transform_reduce(exec, begin(range), end(range), init,
+    std::plus<scalar_type> {},
+    [=](size_type i) { return v1[i] * v2[i]; });
+
+  return sum;
 }
 
 template<class ExecutionPolicy,
@@ -155,7 +164,7 @@ Scalar dot(std::experimental::mdspan<ElementType1, std::experimental::extents<Si
            std::experimental::mdspan<ElementType2, std::experimental::extents<SizeType2, ext2>, Layout2, Accessor2> v2,
            Scalar init)
 {
-  return dot(std::experimental::linalg::impl::default_exec_t(), v1, v2, init);
+  return dot(std::experimental::linalg::impl::default_exec(), v1, v2, init);
 }
 
 template<class ElementType1,
--- a/include/experimental/__p1673_bits/linalg_execpolicy_mapper.hpp
+++ b/include/experimental/__p1673_bits/linalg_execpolicy_mapper.hpp
@@ -12,11 +12,12 @@ inline namespace __p1673_version_0 {
 namespace linalg {
 namespace impl {
 // the execution policy used for default serial inline implementations
-struct inline_exec_t {};
+using inline_exec_t = std::execution::sequenced_policy;
 
 // The execution policy used when no execution policy is provided
 // It must be remapped to some other execution policy, which the default mapper does
-struct default_exec_t {};
+using default_exec_t = std::execution::parallel_policy;
+auto default_exec() { return std::execution::par; };
 
 // helpers
 template<class T> struct is_inline_exec : std::false_type{};
--- a/tests/kokkos-based/matrix_frob_norm_kokkos.cpp
+++ b/tests/kokkos-based/matrix_frob_norm_kokkos.cpp
@@ -87,7 +87,7 @@ TEST_F(blas2_signed_float_fixture, kokkos_matrix_frob_norm_trivial_empty)
 {
   std::vector<value_type> v;
 
-  constexpr auto de = std::experimental::dynamic_extent;
+  constexpr auto de = std::dynamic_extent;
   using s_t = std::experimental::mdspan<value_type, std::experimental::extents<de, de>>;
   s_t M(v.data(), 0, 0);
   namespace stdla = std::experimental::linalg;
@@ -109,7 +109,7 @@ TEST_F(blas2_signed_float_fixture, kokkos_matrix_frob_norm_trivial_one_element)
   constexpr value_type myvalue = static_cast<value_type>(-1.5);
   std::vector<value_type> v = {myvalue};
 
-  constexpr auto de = std::experimental::dynamic_extent;
+  constexpr auto de = std::dynamic_extent;
   using s_t = std::experimental::mdspan<value_type, std::experimental::extents<de, de>>;
   s_t M(v.data(), 1, 1);
   namespace stdla = std::experimental::linalg;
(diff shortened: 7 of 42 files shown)


examples/01_scale.cpp
```diff
@@ -13,7 +13,7 @@
 // Make mdspan less verbose
 using std::experimental::mdspan;
 using std::experimental::extents;
-using std::experimental::dynamic_extent;
+using std::dynamic_extent;
```
mhoemmen, line 16: This change suggests that CI doesn't actually build the examples.  Should we consider fixing that?

This change might actually need to depend on the C++ version, as `std::dynamic_extent` entered the Standard in C++20 with `span`.  We'll have to revisit the examples anyway, because of the recent change to use macros to specify the namespaces.  (The macros let users control them, so that they can but don't need to go into `std`.)

examples/03_matrix_vector_product_mixedprec.cpp
```diff
@@ -5,9 +5,9 @@
 // Make mdspan less verbose
 using std::experimental::mdspan;
 using std::experimental::extents;
-using std::experimental::dynamic_extent;
+using std::dynamic_extent;
 using std::experimental::submdspan;
-using std::experimental::full_extent;
+using std::full_extent;
```
mhoemmen, line 10: Please see above comment on `std::full_extent`; thanks!

include/experimental/__p1673_bits/blas1_dot.hpp
```diff
@@ -43,6 +43,7 @@
 #ifndef LINALG_INCLUDE_EXPERIMENTAL___P1673_BITS_BLAS1_DOT_HPP_
 #define LINALG_INCLUDE_EXPERIMENTAL___P1673_BITS_BLAS1_DOT_HPP_
 
+#include <ranges>
```
mhoemmen, line 46: `<ranges>` is a C++20 include.  Would you consider protecting both the include and the use of `iota_view` below with the appropriate feature test macro, and providing a fall-back implementation?  It's OK if the fall-back is not parallel.

include/experimental/__p1673_bits/blas1_dot.hpp
```diff
@@ -100,10 +101,18 @@ Scalar dot(
...
-  return init;
+  using scalar_type = std::common_type_t<ElementType1, ElementType2, Scalar>;
+  using std::ranges::iota_view;
+  using std::ranges::begin;
+  using std::ranges::end;
+
+  iota_view range{size_type{}, v1.extent(0)};
+
+  Scalar sum = std::transform_reduce(exec, begin(range), end(range), init,
+    std::plus<scalar_type> {},
```
mhoemmen, line 112: Would you consider using the `std::plus<void>` specialization, which deduces argument and return types?  That would fix the `scalar_type` issue above.  It lets the two input types differ, and thus preserves the semantics of the original sequential code.

https://en.cppreference.com/w/cpp/utility/functional/plus_void

```suggestion
    std::plus<void>{},
```

include/experimental/__p1673_bits/blas1_dot.hpp
```diff
@@ -100,10 +101,18 @@ Scalar dot(
                 v1.static_extent(0) == v2.static_extent(0));
 
   using size_type = std::common_type_t<SizeType1, SizeType2>;
-  for (size_type k = 0; k < v1.extent(0); ++k) {
-    init += v1(k) * v2(k);
-  }
-  return init;
+  using scalar_type = std::common_type_t<ElementType1, ElementType2, Scalar>;
```
mhoemmen, line 104: This is not necessary the right type.  For example, if `operator*` returns a higher-precision type than its inputs (as might make sense for custom integer or fixed-point real types, for example), then what you want here is the common type of `Scalar` and the result of `operator*`.  However, it turns out that you don't need `scalar_type` here; please see the comment below.

include/experimental/__p1673_bits/blas1_dot.hpp
```diff
@@ -100,10 +101,18 @@ Scalar dot(
...
+  using std::ranges::iota_view;
+  using std::ranges::begin;
+  using std::ranges::end;
+
+  iota_view range{size_type{}, v1.extent(0)};
+
+  Scalar sum = std::transform_reduce(exec, begin(range), end(range), init,
+    std::plus<scalar_type> {},
+    [=](size_type i) { return v1[i] * v2[i]; });
```
mhoemmen, line 113: This looks correct.  Another approach would be to use `std::views::transform` on the `iota` view to turn `i` into `v1[i]` (or `v2[i]`).  Then you could use `std::plus<void>` directly, instead of creating a lambda that captures `v1` and `v2`.

include/experimental/__p1673_bits/blas1_dot.hpp
```diff
@@ -100,10 +101,18 @@ Scalar dot(
...
-  for (size_type k = 0; k < v1.extent(0); ++k) {
-    init += v1(k) * v2(k);
-  }
-  return init;
+  using scalar_type = std::common_type_t<ElementType1, ElementType2, Scalar>;
+  using std::ranges::iota_view;
+  using std::ranges::begin;
+  using std::ranges::end;
+
+  iota_view range{size_type{}, v1.extent(0)};
```
mhoemmen, line 109: The preferred way to use range factories is `std::views::FOO` (in this case, `std::views::iota`) instead of `std::ranges::FOO_view`.

include/experimental/__p1673_bits/blas1_dot.hpp
```diff
@@ -100,10 +101,18 @@ Scalar dot(
...
+
+  Scalar sum = std::transform_reduce(exec, begin(range), end(range), init,
```
mhoemmen, line 111: The "inline" executor should be executing inline.  This means that, while it can use Standard Library algorithms, it shouldn't be passing along any execution policy (even if that execution policy is "sequential" -- please see my comments below).  Would you consider changing this to remove the execution policy argument?

```suggestion
  Scalar sum = std::transform_reduce(begin(range), end(range), init,
```

include/experimental/__p1673_bits/linalg_execpolicy_mapper.hpp
```diff
@@ -12,11 +12,12 @@ inline namespace __p1673_version_0 {
 namespace linalg {
 namespace impl {
 // the execution policy used for default serial inline implementations
-struct inline_exec_t {};
+using inline_exec_t = std::execution::sequenced_policy;
```
mhoemmen, line 15: Please see note above about the "inline" executor.  

Any Standard Algorithm overload that takes an `ExecutionPolicy` template parameter is a "parallel algorithm" (see [[algorithms.parallel] 2](https://eel.is/c++draft/algorithms.parallel#defns-2)).  That changes what the algorithm assumes about the behavior of element access functions. 

Therefore, it's really important that the "inline" executor not use any parallel algorithm, even if `ExecutionPolicy` is `sequenced_policy`.

include/experimental/__p1673_bits/linalg_execpolicy_mapper.hpp
```diff
@@ -12,11 +12,12 @@ inline namespace __p1673_version_0 {
...
 
 // The execution policy used when no execution policy is provided
 // It must be remapped to some other execution policy, which the default mapper does
-struct default_exec_t {};
+using default_exec_t = std::execution::parallel_policy;
```
mhoemmen, line 19: `// It must be remapped to some other execution policy, which the default mapper does` -- it's important that we not circumvent the default mapper.  Would you consider, thus, reverting this change?

tests/kokkos-based/matrix_frob_norm_kokkos.cpp
```diff
@@ -87,7 +87,7 @@ TEST_F(blas2_signed_float_fixture, kokkos_matrix_frob_norm_trivial_empty)
 {
   std::vector<value_type> v;
 
-  constexpr auto de = std::experimental::dynamic_extent;
+  constexpr auto de = std::dynamic_extent;
```
mhoemmen, line 90: Does CI not actually test this test?  (Please see comment above about `std::dynamic_extent`.)  If so, then should we consider using the namespace macros instead of `std` explicitly?

tests/native/dot.cpp
```diff
@@ -74,14 +75,13 @@ namespace {
...
-    // This is noncomforming, but I need some way to test the executor overloads.
-    //using fake_executor_t = int;
-    //dot (fake_executor_t (), x, y, dotResultPar);
-    //EXPECT_EQ( dotResultPar, expectedDotResult );
+#ifdef LINALG_ENABLE_TBB
+    const auto dotResultPar = dot (std::execution::par, x, y);
+#else
+    const auto dotResultPar = dot (std::execution::seq, x, y);
+#endif // LINALG_ENABLE_TBB
+    static_assert( std::is_same_v<std::remove_const_t<decltype(dotResultPar)>, scalar_t> );
```
mhoemmen, line 83:
```suggestion
#ifdef LINALG_ENABLE_TBB
    auto dotResultPar = dot (std::execution::par, x, y);
#else
    auto dotResultPar = dot (std::execution::seq, x, y);
#endif // LINALG_ENABLE_TBB
    static_assert( std::is_same_v<decltype(dotResultPar), scalar_t> );
```

CMakeLists.txt
```diff
@@ -130,6 +130,15 @@ if(LINALG_ENABLE_KOKKOS)
   find_package(KokkosKernels REQUIRED)
 endif()
 
+find_package(TBB)
```
mhoemmen, line 133: More recent versions of GCC (for instance) shouldn't require TBB for `std::execution::par` to work.  Furthermore, nvc++ comes with its own, non-TBB implementation of `std::execution::par`.  Would you consider gating this on compiler version, instead of requiring a third-party library?  That could even be done in the code -- the code would just need to test the appropriate compiler version to ensure that the C++ algorithms are available (see https://en.cppreference.com/w/cpp/compiler_support/17 and search for "Parallel algorithms and execution policies").

mhoemmen (changes requested):
Thanks for this contribution! : - )  Please see comments; thanks!

examples/01_scale.cpp
amklinv-nnl: CI builds the examples; I just wasn't building them on my PC.

I was unaware of the macros change. Can you give me an example?

include/experimental/__p1673_bits/blas1_dot.hpp
amklinv-nnl: Sure! Can you point me to an example of using a feature test macro?

include/experimental/__p1673_bits/blas1_dot.hpp
amklinv-nnl: Thanks for the input. Can you show me what that would look like?

include/experimental/__p1673_bits/blas1_dot.hpp
amklinv-nnl: Sure. What do we need to add to handle other execution policies?

include/experimental/__p1673_bits/linalg_execpolicy_mapper.hpp
amklinv-nnl: If I remember correctly, the execpolicy_mapper did not like this default. I can restore it and show you the error it gave me.

examples/01_scale.cpp
mhoemmen: `MDSPAN_IMPL_STANDARD_NAMESPACE` and `MDSPAN_IMPL_PROPOSED_NAMESPACE` are the two macros.  They can be defined by users, but they also get default definitions, e.g., [here in `include/experimental/mdspan`](https://github.com/kokkos/mdspan/blob/640ca0166589bc9be7dbbeecc62c4f5edad2915c/include/experimental/mdspan#L19).  The library assumes that `MDSPAN_IMPL_PROPOSED_NAMESPACE` is nested inside `MDSPAN_IMPL_STANDARD_NAMESPACE`.

Using these macros might require updating the version of the reference mdspan implementation that github's CI pulls in.

include/experimental/__p1673_bits/blas1_dot.hpp
mhoemmen: Sure!  [This web page](https://en.cppreference.com/w/cpp/utility/feature_test) gives a good summary.

```c++
// Ensure that the feature test macro __cpp_lib_ranges is available;
// <version> also defines this macro, but that is a C++20 header.
#include <algorithm>

#if defined(__cpp_lib_ranges)
#  include <ranges>
#endif

void some_function() {
#if defined(__cpp_lib_ranges_iota)
  // ... code using views::iota ...
#else
  // ... fall-back code ...
#endif
}
```

The point of using two different macros -- one for the header, and one for the specific feature `iota::view` -- is that the feature came after the header, so some compiler versions may have the header but not the feature.

examples/02_matrix_vector_product_basic.cpp
```diff
@@ -31,11 +31,11 @@ int main(int argc, char* argv[]) {
     mdspan<double, extents<std::size_t, dynamic_extent>> y(y_vec.data(),N);
     for(int i=0; i<A.extent(0); i++)
       for(int j=0; j<A.extent(1); j++)
-        A(i,j) = 100.0*i+j;
+        A[i,j] = 100.0*i+j;
```
mhoemmen, line 34: This will only compile if the C++23 feature "multidimensional subscript operator" (P2128R6) is available.  As a result, would you consider one of the following changes?

1. Protect the example so it only compiles if the feature test macro `__cpp_multidimensional_subscript` is defined, OR
2. Add `#define MDSPAN_USE_PAREN_OPERATOR 1` to the example before including any `mdspan` headers (see top of https://godbolt.org/z/Yrr8oe9sE for an example of the relevant macros), and use parentheses instead of brackets (e.g., `A(i,j)`)

include/experimental/__p1673_bits/blas1_linalg_add.hpp
```diff
@@ -132,7 +132,7 @@ void add_rank_2(
   using size_type = std::common_type_t<SizeType_x, SizeType_y, SizeType_z>;
   for (size_type j = 0; j < x.extent(1); ++j) {
     for (size_type i = 0; i < x.extent(0); ++i) {
-      z(i,j) = x(i,j) + y(i,j);
+      z[i,j] = x[i,j] + y[i,j];
```
mhoemmen, line 135: The actual implementation is a C++17 back-port, so we unfortunately have to roll with whatever operator (parentheses or brackets) the user has available.  If you really do want to change the implementation, then we might have to go with something like the following.

```suggestion
#if (MDSPAN_USE_PAREN_OPERATOR > 0)
            z(i,j) = x(i,j) + y(i,j);
#else
            z[i,j] = x[i,j] + y[i,j];
#endif
```

or at least

```suggestion
#if defined(__cpp_multidimensional_subscript)
            z[i,j] = x[i,j] + y[i,j];
#else
            z(i,j) = x(i,j) + y(i,j);
#endif
```

It might be best just to leave the implementation alone; we might want to come up with a better way to do this.

mhoemmen (changes requested):
I would recommend, at least for now, not changing the implementation to use the multidimensional subscript operator.  Doing this would hinder back-porting to C++17, which is an important requirement of the reference implementation.

PR #253, opened 2023-06-13
Fix C++23 build (GCC 13.1.0)

mhoemmen:
Fix tests and examples to build with C++23, using GCC 13.1.0.

I set the following CMake options.

```bash
cmake ${SOME_SOURCE_DIRECTORY_ROOT}/stdBLAS \
  -DCMAKE_CXX_COMPILER=${PATH_TO_LOCAL_GCC}/bin/g++ \
  -DCMAKE_C_COMPILER=${PATH_TO_LOCAL_GCC}/bin/gcc \
  -DLINALG_ENABLE_TESTS=ON \
  -DLINALG_ENABLE_EXAMPLES=ON \
  -DLINALG_CXX_STANDARD=23
```

where `SOME_SOURCE_DIRECTORY_ROOT` is the parent directory of my stdBLAS source checkout, and `PATH_TO_LOCAL_GCC` is the installation directory of my local GCC 13.1.0 build.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -102,8 +102,6 @@ jobs:
         
   test-stdBLAS:
     runs-on: ubuntu-latest
-    container:
-      image: amklinv/mdspan-dependencies:latest
     needs: build-stdblas
     
     steps:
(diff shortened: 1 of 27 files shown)


mhoemmen:
The test appears to be failing because it can't find GCC's shared libraries.  @amklinv-nnl would you happen to know about this?  (Thanks! : - ) )

amklinv-nnl:
> The test appears to be failing because it can't find GCC's shared libraries. @amklinv-nnl would you happen to know about this? (Thanks! : - ) )

I sure would! It's because the test is trying to use a container with a different gcc version. My PR fixes that problem.

PR #254, opened 2023-06-14
Fix TBB build issue

amklinv-nnl:
@mhoemmen Please review #252 first since this PR sits on top of it.

PR #256, opened 2023-06-27
Add conditional stdpar feature

mhoemmen:
GCC, even 13.1.0, needs TBB in order for `<execution>` features to compile.  Thus, we added a `LINALG_ENABLE_TBB` CMake option, and documented the option in README.md.  We've tested the option using a local TBB from-source build from their main branch, with GCC 13.1.0.  (You may need to set the CMake option `TBB_STRICT=OFF`, to disable warnings as errors in TBB.)

Previously, stdBLAS was using the `MDSPAN_EXAMPLES_USE_EXECUTION_POLICIES` macro to protect use of `std::execution` features in examples.  We added a new macro `LINALG_HAS_EXECUTION` (conditionally defined in macros.hpp) to indicate that `std::execution` features can be used, and replaced all instances of `MDSPAN_EXAMPLES_USE_EXECUTION_POLICIES` with `LINALG_HAS_EXECUTION`.

While doing this, we noticed that macros.hpp was not including the CMake-generated header file `linalg_config.h`, even though `linalg_config.h` has a nonzero number of macro definitions in it.  Thus, we made sure that macros.hpp includes `linalg_config.h`, so that macros.hpp (and thereby all stdBLAS headers) can pick up CMake options.  (Otherwise, CMake options aren't very useful!)

Requiring a CMake-generated header file effectively makes stdBLAS no longer a header-only library.  We consider this reasonable, especially since many implementations in practice would want to link against a binary-only BLAS library.  The only other option would be to require GCC users to use TBB.  However, the TBB repository doesn't seem to have software releases past 2021, and it's not clear whether the main branch can be considered stable.  Thus, we think conditionally disabling `<execution>` features if using GCC and TBB is not available is reasonable.

These changes were copied over from PR #254 (thanks to @amklinv-nnl for implementing them!).  This PR supersedes PR #254.


crtrott (approved):

mhoemmen:
@crtrott fixed issues with this PR and submitted that as a new PR https://github.com/kokkos/stdBLAS/pull/283 .  PR #283 replaces this PR.

PR #257, opened 2023-06-29
Make scale work for std::execution::par

mhoemmen:
Make `scale` work for `std::execution::par`.

This is rebased atop PR #256; please merge that first; thanks!


mhoemmen:
@amklinv-nnl this could be helpful for informing your parallelization efforts

PR #260, opened 2023-07-26
Bring conjugated_accessor and scaled_accessor in line with the C++ Working Draft

mhoemmen:
* Rename `accessor_scaled` to `scaled_accessor`
* Change `scaled_accessor::reference` to the actual product value, and change `element_type` accordingly
* Add full converting constructor (scaling factor and/or input `scaled_accessor`'s nested accessor could differ)
* Make `scaled_accessor::access` conditionally (instead of unconditionally) `noexcept`
* Rename `accessor_conjugate` to `conjugate_accessor`
* Implement analogous changes to `conjugate_accessor`
* Fix implementation of _`conj-if-needed`_, and add implementations (and tests) of _`real-if-needed`_ and _`imag-if-needed`_
* Add tests mixing result of `scaled` with an mdspan whose accessor uses a proxy reference as its `reference` type.  This increases confidence that the new (proxy-reference-free) design works.


include/experimental/__p1673_bits/conjugated.hpp
```diff
@@ -51,74 +51,34 @@ inline namespace __p1673_version_0 {
...
 private:
   Accessor accessor_;
-  using aliases = impl::accessor_conjugate_aliases<Accessor>;
 
 public:
-  using reference        = typename aliases::reference;
-  using element_type     = typename aliases::element_type;
-  using data_handle_type = typename aliases::data_handle_type;
-  using offset_policy    = typename aliases::offset_policy;
+  using element_type = decltype(impl::conj_if_needed(std::declval<typename Accessor::element_type>()));
```
mhoemmen, line 59: We need `add_const_t` here, as making the `element_type` const would emphasize that this accessor is always read only.

mhoemmen: There's another issue here, which is that for the wording change in https://github.com/ORNL/cpp-proposals-pub/pull/401 , the definition of `element_type` is not consistent with the definition of `access`.  I've commented on this below.

crtrott (approved):

crtrott:
I counter checked with standard, and compiled and ran tests.

Issue #267, opened 2024-03-27
Implement paper that adds transposed special cases for P2642 layouts


mhoemmen:
See https://github.com/ORNL/cpp-proposals-pub/pull/454 (paper PR).

mhoemmen:
P3222 is that paper; I submitted it and LEWG plans to review it soon.

PR #268, opened 2024-03-27
Implement P3222 and P3050

mhoemmen:
Implement P3222R0 ("Add `transposed` special cases for P2642 layouts").  The corresponding paper PR is https://github.com/ORNL/cpp-proposals-pub/issues/448.  Add tests for previously supported cases and the new cases.

Implement P3050R2 ("Optimize `linalg::conjugated` for noncomplex value types") and add tests.  That is, fix `conjugated` for non-arithmetic, non-(custom complex) types.  A type `T` is "custom complex" if `conj(T)` is ADL-findable.

Each of these papers has a separate CMake option, which is documented in CMakeLists.txt.  Both options default to `OFF` (*not* implemented) for now.

Fixes https://github.com/kokkos/stdBLAS/issues/267 .


examples/01_scale.cpp
```diff
@@ -16,13 +17,16 @@
 #  include <execution>
 #endif
 
-using std::experimental::mdspan;
-using std::experimental::extents;
+namespace MdSpan = MDSPAN_IMPL_STANDARD_NAMESPACE;
+namespace LinearAlgebra = MDSPAN_IMPL_STANDARD_NAMESPACE :: MDSPAN_IMPL_PROPOSED_NAMESPACE :: linalg;
```
dalg24, line 21: Obviously not a blocker but why did we not define `MDSPAN_IMPL_PROPOSED_NAMESPACE` to be `Experimental -> Kokkos::Experimental` or `experimental -> std::experimental` so we don't have to nest it with `MDSPAN_IMPL_STANDARD_NAMESPACE`?
Also I can't remember why we decided to use macros rather than alias namespaces.

dalg24 (commented):
I would prefer if you did not mix in the refactor/fixes with the implementation of the new feature.

examples/01_scale.cpp
mhoemmen: Thanks @dalg24 !  I discussed this question with @crtrott a while back.  I recall that he thought it would be a lot more work to assume that the two namespaces aren't nested.

mhoemmen:
Hi @dalg24 !  Thanks for your review!

> I would prefer if you did not mix in the refactor/fixes with the implementation of the new feature.

It's actually impossible to pass the repository's automated premerge tests without the fixes, as the build fails.

dalg24:
Can't you open another PR with the fixes only?

mhoemmen:
> Can't you open another PR with the fixes only?

There are lots of fixes.  They are separated into different commits.

The current state of the repo is broken; it fails to build.

mhoemmen:
@dalg24 Per your request, I've created PR #269 that only fixes the build and Standard conformance issues, without adding new features.

This PR is rebased atop PR #269, because (as mentioned before) this repository's build is currently broken, so it's impossible to pass check-in tests without the build fixes.  Please merge PR #269 first.

mhoemmen:
@crtrott @dalg24 Changes as of today:

1. Dependency PR #269 has merged.
2. Added separate CMake options to enable each paper's changes.
3. CMake options default to `OFF` (paper's changes *not* enabled) by default.

crtrott (approved):
This looks good. I am not 100% sure I like the approach for the conjugated thing but its anyway protected so I am good.

include/experimental/__p1673_bits/conjugated.hpp
```diff
@@ -94,10 +94,21 @@ class conjugated_accessor {
 template<class ElementType, class Extents, class Layout, class Accessor>
 auto conjugated(mdspan<ElementType, Extents, Layout, Accessor> a)
 {
-  if constexpr (std::is_arithmetic_v<std::remove_cv_t<ElementType>>) {
+  using value_type = typename decltype(a)::value_type;
+
+  if constexpr (std::is_arithmetic_v<value_type>) {
```
iburyl, line 99: #ifdef should be somehwhere here, p1673 spec has nothing about arithmetic types. Only else branch is valid in 1673 context

iburyl:
Other than small comment on scope of #ifdef looks fine to me

mhoemmen:
@crtrott wrote:
> This looks good. I am not 100% sure I like the approach for the conjugated thing but its anyway protected so I am good.

Just to clarify: P3050 is an optimization that simplifies the implementation.  It needs some way to tell if a number type is _not_ complex.  It can be conservative about that test, because it's only an optimization.  Implementations can always optimize internally by specializing on known `conjugated_accessor` patterns.

This is different than P3371R1.  As we discussed yesterday, P3371R1 changes Hermitian rank-1 and rank-k updates to constrain `Scalar` _not_ to have ADL-findable `conj`, even if the user's `Scalar` is a "noncomplex" number type.  We talked yesterday about changing this in P3371R2 from a Constraint (the user's code fails to build) to a Precondition that _`imag-if-needed`_`(alpha)` equals `Scalar{}`.

include/experimental/__p1673_bits/conjugated.hpp
mhoemmen: Good point!  I'll fix this.

mhoemmen: I just pushed a fix.  The tests now cover this case.

PR #275, opened 2024-07-05
Fix rank_k implementation

### LINKED ISSUE #272: symmetric_matrix_rank_k_update != dsyrk
As mentioned at the end of #261, I tried the example posted by @yigiter in https://github.com/kokkos/stdBLAS/issues/261#issuecomment-1873645288_ which lead me to wonder what `symmetric_matrix_rank_k_update` is computing.

Given its name, I assume that  `symmetric_matrix_rank_k_update` maps to the BLAS function [`dsyrk`](https://netlib.org/lapack/explore-html-3.6.1/d1/d54/group__double__blas__level3_gae0ba56279ae3fa27c75fefbc4cc73ddf.html) (but with a single scalar constant alpha, rather than BLAS's alpha and beta) described as:

>  DSYRK  performs one of the symmetric rank k operations
>
>    C := alpha*A*A**T + beta*C,
>
> or
>
>    C := alpha*A**T*A + beta*C,
>
> where  alpha and beta  are scalars, C is an  n by n  symmetric matrix
> and  A  is an  n by k  matrix in the first case and a  k by n  matrix
> in the second case.

I reused @yigiter's example as follows:

```
#define MDSPAN_USE_PAREN_OPERATOR 1

#include <type_traits>
#include <mdspan/mdspan.hpp>
#include <experimental/linalg>
#include <fmt/ranges.h>
extern "C" {
        #include <cblas.h>
}

using namespace MDSPAN_IMPL_STANDARD_NAMESPACE;
using namespace MDSPAN_IMPL_STANDARD_NAMESPACE :: MDSPAN_IMPL_PROPOSED_NAMESPACE;
using namespace MDSPAN_IMPL_STANDARD_NAMESPACE :: MDSPAN_IMPL_PROPOSED_NAMESPACE :: linalg;

void print(auto const& M)
{
        for (int i = 0; i < M.extent(0); ++i) {
                for (int j = 0; j < M.extent(1); ++j) {
                        fmt::print("{:3} ", M(i,j));
                }
                fmt::print("\n");
        }
}

int main()
{
        double lgAvec[6]={1,1,1,2,2,2};
        double lgCvec[9]={1,2,3,2,4,5,3,5,6};
        mdspan lgA(lgAvec,3,2);
        mdspan lgC(lgCvec,3,3);
        fmt::print("============== linalg ==============\n");
        fmt::print("============== before ==============\n");
        fmt::print("A=\n");print(lgA);
        fmt::print("C=\n");print(lgC);
        symmetric_matrix_rank_k_update(lgA, lgC, upper_triangle);
        fmt::print("============== after  ==============\n");
        fmt::print("A=\n");print(lgA);
        fmt::print("C=\n");print(lgC);

        fmt::print("============== openblas ============\n");
        double blAvec[6]={1,1,1,2,2,2};
        double blCvec[9]={1,2,3,2,4,5,3,5,6};
        mdspan blA(blAvec,3,2);
        mdspan blC(blCvec,3,3);
        fmt::print("============== before ==============\n");
        fmt::print("A=\n");print(blA);
        fmt::print("C=\n");print(blC);
        cblas_dsyrk(CblasRowMajor, CblasUpper, CblasNoTrans, 3, 2, 1.0, blAvec, 2, 1.0, blCvec, 3);
        fmt::print("============== after  ==============\n");
        fmt::print("A=\n");print(blA);
        fmt::print("C=\n");print(blC);
}
```
which generates the output:
```
============== linalg ==============
============== before ==============
A=
  1   1 
  1   2 
  2   2 
C=
  1   2   3 
  2   4   5 
  3   5   6 
============== after  ==============
A=
  1   1 
  1   2 
  2   2 
C=
  4   6   3 
  6  12   5 
  3   5   6 
============== openblas ============
============== before ==============
A=
  1   1 
  1   2 
  2   2 
C=
  1   2   3 
  2   4   5 
  3   5   6 
============== after  ==============
A=
  1   1 
  1   2 
  2   2 
C=
  3   5   7 
  2   9  11 
  3   5  14 
```
I expect the answer
```
    3    5    7
    5    9   11
    7   11   14
```
as per the blas function (upper triangle).

Given some matrix `A`, `A A^T` is going to be symmetric, so the symmetric constraint needs to be on the additive matrix `C` in order to avoid a triangle of computations, The comment in https://github.com/kokkos/stdBLAS/blob/06e90a58b67c5adefff0f06904c8f8bc3371815b/include/experimental/__p1673_bits/blas3_matrix_rank_k_update.hpp#L152 seems to suggest that `A` is the matrix which needs to be symmetric?

I am puzzled as to what `symmetric_matrix_rank_k_update` is meant to be computing?

rasolca:
The implementation of `symmetric_matrix_rank_k_update` and `hermitian_matrix_rank_k_update` are wrong.

See xSYRK or xHERK for reference.
(E.g. [DSYRK](https://netlib.org/lapack/explore-html-3.6.1/d1/d54/group__double__blas__level3_gae0ba56279ae3fa27c75fefbc4cc73ddf.html))

Fixes #272

### DIFF
--- a/include/experimental/__p1673_bits/blas3_matrix_rank_k_update.hpp
+++ b/include/experimental/__p1673_bits/blas3_matrix_rank_k_update.hpp
@@ -147,8 +147,6 @@ void symmetric_matrix_rank_k_update(
   mdspan<ElementType_C, extents<SizeType_C, numRows_C, numCols_C>, Layout_C, Accessor_C> C,
   Triangle /* t */)
 {
-  using size_type = std::common_type_t<SizeType_A, SizeType_C>;
-
   constexpr bool lower_tri =
     std::is_same_v<Triangle, lower_triangle_t>;
   using size_type = std::common_type_t<SizeType_A, SizeType_C>;
--- a/include/experimental/__p1673_bits/blas3_matrix_rank_k_update.hpp
+++ b/include/experimental/__p1673_bits/blas3_matrix_rank_k_update.hpp
@@ -149,31 +149,16 @@ void symmetric_matrix_rank_k_update(
 {
   using size_type = std::common_type_t<SizeType_A, SizeType_C>;
 
-  if constexpr (std::is_same_v<Triangle, upper_triangle_t>) {
-    // [A_00  A_01  A_02]   [A_00            ]
-    // [      A_11  A_12] * [A_01  A_11      ]
-    // [            A_22]   [A_02  A_12  A_22]
-    for (size_type row = 0; row < C.extent(0); ++row) {
-      for (size_type col = 0; col < C.extent(1); ++col) {
-        const size_type lower = row < col ? col : row; // max
-        const size_type upper = A.extent(0);
-        for (size_type k = lower; k < upper; ++k) {
-          C(row, col) += alpha * A(row, k) * A(col, k);
-        }
-      }
-    }
-  }
-  else {
-    // [A_00            ]   [A_00  A_10  A_20]
-    // [A_10  A_11      ] * [      A_11  A_21]
-    // [A_20  A_21  A_22]   [            A_22]
-    for (size_type row = 0; row < C.extent(0); ++row) {
-      for (size_type col = 0; col < C.extent(1); ++col) {
-        const size_type lower = 0;
-        const size_type upper = row < col ? row : col; // min
-        for (size_type k = lower; k < upper; ++k) {
-          C(row, col) += alpha * A(row, k) * A(col, k);
-        }
+  constexpr bool lower_tri =
+    std::is_same_v<Triangle, lower_triangle_t>;
+  using size_type = std::common_type_t<SizeType_A, SizeType_C>;
+
+  for (size_type j = 0; j < C.extent(1); ++j) {
+    const size_type i_lower = lower_tri ? j : size_type(0);
+    const size_type i_upper = lower_tri ? C.extent(0) : j+1;
+    for (size_type i = i_lower; i < i_upper; ++i) {
+      for (size_type k = 0; k < A.extent(1); ++k) {
+          C(i, j) += alpha * A(i, k) * A(j, k);
       }
     }
   }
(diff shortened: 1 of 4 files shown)


mhoemmen:
Thanks for contributing!  We'll review this as soon as we can.

include/experimental/__p1673_bits/blas3_matrix_rank_k_update.hpp
```diff
@@ -149,31 +149,16 @@ void symmetric_matrix_rank_k_update(
...
-    for (size_type row = 0; row < C.extent(0); ++row) {
-      for (size_type col = 0; col < C.extent(1); ++col) {
-        const size_type lower = 0;
-        const size_type upper = row < col ? row : col; // min
-        for (size_type k = lower; k < upper; ++k) {
-          C(row, col) += alpha * A(row, k) * A(col, k);
-        }
+  constexpr bool lower_tri =
+    std::is_same_v<Triangle, lower_triangle_t>;
+  using size_type = std::common_type_t<SizeType_A, SizeType_C>;
```
mhoemmen, line 154: Line 154 is redundant with line 150.

```suggestion
  constexpr bool lower_tri =
    std::is_same_v<Triangle, lower_triangle_t>;
```

include/experimental/__p1673_bits/blas3_matrix_rank_k_update.hpp
```diff
@@ -149,31 +149,16 @@ void symmetric_matrix_rank_k_update(
 {
   using size_type = std::common_type_t<SizeType_A, SizeType_C>;
```
mhoemmen, line 151: This type alias is redundant with line 154.

```suggestion
```

mhoemmen (approved):
Not sure what we were thinking there -- thank you for the fix! : - )

mhoemmen:
I'm able to reproduce the `static_assert` locally on the main branch.  I'll fix that first.

PR #276, opened 2024-07-18
Fix transposed test

mhoemmen:
Test did not account for `submdspan` of `layout_right` being able to return `layout_right_padded` instead of `layout_stride`.


mhoemmen:
I just pushed the commit to PR https://github.com/kokkos/stdBLAS/pull/275 .

PR #282, opened 2024-08-27
read only real part from the diagonal of the hermitian matrix (part 2)

iburyl:
Another PR (https://github.com/kokkos/stdBLAS/pull/278) fixed (among other things) reading only real part of the diagonal for hermitian matrix-vector product.

This PR fixes the same thing for the matrix-matrix product.

### DIFF
--- a/include/experimental/__p1673_bits/blas3_matrix_product.hpp
+++ b/include/experimental/__p1673_bits/blas3_matrix_product.hpp
@@ -1737,9 +1737,12 @@ void hermitian_matrix_left_product(
     for (size_type j = 0; j < C.extent(1); ++j) {
       for (size_type i = 0; i < C.extent(0); ++i) {
         C(i,j) = ElementType_C{};
-        for (size_type k = 0; k < A.extent(1); ++k){
-          ElementType_A aik = i <= k ? impl::conj_if_needed(A(k,i)) : A(i,k);
-          C(i,j) += aik * B(k,j);
+        for (size_type k = 0; k < i; ++k){
+          C(i,j) += A(i,k) * B(k,j);
+        }
+        C(i,j) += impl::real_part(A(i,i)) * B(i,j);
+        for (size_type k = i+1; k < A.extent(0); ++k){
+          C(i,j) += impl::conj_if_needed(A(k,i)) * B(k,j);
         }
       }
     }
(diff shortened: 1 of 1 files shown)


include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -1737,9 +1737,12 @@ void hermitian_matrix_left_product(
...
-        for (size_type k = 0; k < A.extent(1); ++k){
-          ElementType_A aik = i <= k ? impl::conj_if_needed(A(k,i)) : A(i,k);
-          C(i,j) += aik * B(k,j);
+        for (size_type k = 0; k < i; ++k){
+          C(i,j) += A(i,k) * B(k,j);
+        }
+        C(i,j) += impl::real_part(A(i,i)) * B(i,j);
+        for (size_type k = i+1; k < A.extent(0); ++k){
+          C(i,j) += impl::conj_if_needed(A(k,i)) * B(k,j);
         }
```
mhoemmen, line 1746: It should be legal for `A.extent(0)` to be zero, meaning that `A` has no diagonal element.  Should we thus consider the following change?

```suggestion
        for (size_type k = 0; k < i; ++k){
          C(i,j) += A(i,k) * B(k,j);
        }
        if (A.extent(0) != 0) {
          C(i,j) += impl::real_part(A(i,i)) * B(i,j);
        }
        for (size_type k = i+1; k < A.extent(0); ++k){
          C(i,j) += impl::conj_if_needed(A(k,i)) * B(k,j);
        }
```

include/experimental/__p1673_bits/blas3_matrix_product.hpp
```diff
@@ -1748,9 +1751,12 @@ void hermitian_matrix_left_product(
...
-        for (size_type k = 0; k < A.extent(1); ++k) {
-          ElementType_A aik = i >= k ? impl::conj_if_needed(A(k,i)) : A(i,k);
-          C(i,j) += aik * B(k,j);
+        for (size_type k = 0; k < i; ++k) {
+          C(i,j) += impl::conj_if_needed(A(k,i)) * B(k,j);
+        }
+        C(i,j) += impl::real_part(A(i,i)) * B(i,j);
+        for (size_type k = i+1; k < A.extent(1); ++k) {
+          C(i,j) += A(i,k) * B(k,j);
         }
```
mhoemmen, line 1760:
```suggestion
        for (size_type k = 0; k < i; ++k) {
          C(i,j) += impl::conj_if_needed(A(k,i)) * B(k,j);
        }
        if (A.extent(1) != 0) {
          C(i,j) += impl::real_part(A(i,i)) * B(i,j);
        }
        for (size_type k = i+1; k < A.extent(1); ++k) {
          C(i,j) += A(i,k) * B(k,j);
        }
```

mhoemmen (changes requested):
Thanks for the fixes!  I made a couple suggestions.

include/experimental/__p1673_bits/blas3_matrix_product.hpp
iburyl: Well... `A` is square by precondition. That means `A.extent(0) == A.extent(1) == 0`. And... `C.extent(0) == 0` (via multipliable precondition). We should never find oursevles in that part of the loop in this case -> this is a redundant if condition.

mhoemmen: That's a good point.  While we should consider adding some assert s in there for the precondition, that's out of scope of the PR.

include/experimental/__p1673_bits/blas3_matrix_product.hpp
mhoemmen: Please see above conversation -- thanks!

mhoemmen (approved):

PR #286, opened 2024-09-19
Simplify workflow by not uploading/downloading as much

crtrott:


mhoemmen (approved):
FYI to @amklinv-nnl who worked on this a while back -- it looks like the repo itself can do a lot of this work, so we don't have to put it in the workflow any more.  Thanks! : - )

PR #287, opened 2024-09-26
aligned vector_abs_sum implementation with spec

iburyl:
1) using abs(real) + abs(imag) for non-arithmetic types
2) following GENERALIZED_SUM definition for intermediate calculation types

### DIFF
--- a/include/experimental/__p1673_bits/blas1_vector_abs_sum.hpp
+++ b/include/experimental/__p1673_bits/blas1_vector_abs_sum.hpp
@@ -84,20 +84,28 @@ Scalar vector_abs_sum(
   mdspan<ElementType, extents<SizeType, ext0>, Layout, Accessor> v,
   Scalar init)
 {
-  using std::abs;
-  using result_t = decltype(init + abs(impl::real_part(std::declval<ElementType>())));
+  using value_type = typename decltype(v)::value_type;
+  using sum_type =
+    decltype(init +
+             impl::abs_if_needed(impl::real_part(std::declval<value_type>())) +
+             impl::abs_if_needed(impl::imag_part(std::declval<value_type>())));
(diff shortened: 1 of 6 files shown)


include/experimental/__p1673_bits/blas1_vector_abs_sum.hpp
```diff
@@ -84,12 +84,20 @@ Scalar vector_abs_sum(
   mdspan<ElementType, extents<SizeType, ext0>, Layout, Accessor> v,
   Scalar init)
 {
+  using std::abs;
+  using result_t = decltype(init + abs(impl::real_part(std::declval<ElementType>())));
   const SizeType numElt = v.extent(0);
+  result_t value = init;
```
mhoemmen, line 90: This looks like a good interpretation of [linalg.algs.blas1.asum] 2.  It doesn't quite implement [linalg.algs.blas1.asum] 4, but we didn't implement that correctly before, so it can wait for another PR.

mhoemmen (approved):

include/experimental/__p1673_bits/blas1_vector_abs_sum.hpp
iburyl: [linalg.algs.blas1.asum] 4 says, that if Scalar is of higher precision, than "intermediate terms in the sum use Scalar's precision or greater".

My interpretation: if Scalar is of higher precision, then result_t will be of higher precision.
Thus += something to result_t will be intermediate terms in that higher precision.

This is why I split those, for both to be in that higher precision
```
        value += abs(impl::real_part(v(i)));
        value += abs(impl::imag_part(v(i)));
```

What is wrong with "doesn't quite implement [linalg.algs.blas1.asum] 4"?

P.S. I'll fix the broken test.

PR #288, opened 2024-09-26
Fix hermitian rank-1/2/k/2k updates

iburyl:
Imaginary part of the diagonal is zeroed.
Aligned with BLAS.
Can be considered aligned with spec change, where diagonal is not touched on read (but replaced on store).


mhoemmen (approved):
Thanks for the fix! : - )

PR #290, opened 2024-09-28
Fix vector idx abs max

iburyl:
1) renamed idx_abs_max -> vector_idx_abs_max
2) aligned behavior for complex values with the spec (that is abs is not really abs)
3) Currently if NaN is in the first position output is 0. Output will ignore NaNs in any other position and will provide an index of the max value, ignoring all other NaNs. NaNs are checked the same way as in BLAS (that is not really checked). NaN behavior is not clarified with spec. Anything beyond that implementation will have cross-board performance implications. Leave NaNs to unspecified output.

### DIFF
--- a/examples/kokkos-based/idx_abs_max_kokkos.cpp
+++ b/examples/kokkos-based/idx_abs_max_kokkos.cpp
@@ -11,10 +11,10 @@ void run_trivial_example()
   using extents_type = stdexp::extents<stdexp::dynamic_extent>;
   stdexp::mdspan<value_type, extents_type> a(arr.data(),0);
 
-  const auto idx = stdla::idx_abs_max(std::execution::seq, a);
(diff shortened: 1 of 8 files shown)


crtrott (changes requested):
Looks good to me except that I think the complex test should fail now or? Because `std::abs(z)` is different from `abs(z.real()) + abs(z.imag())` isn't it? Blocking to see what the tests do.

crtrott:
I pushed a change to the test which makes it sensitive to the changed behavior for complex. It wasn't before (i.e. the test passed before and with your change).

crtrott (approved):
Good now. I fixed the test to be sensitive to this change.

crtrott:
@mhoemmen if you also approve I merge (or you can merge and we consider that approval)

PR #291, opened 2024-09-30
Fix for vector_sum_of_squares and renaming for vector_norm2

iburyl:
1. implementation for vector_sum_of_squares was completely broken - fixed
2. vector_norm2 renamed to vecotr_two_norm (as it was done in p1673r12)


iburyl:
sequencies in tests are monotonically increasing in abs values, thus broken branch is never touched

mhoemmen (approved):
Thanks Ilya! : - )

PR #292, opened 2024-09-30
renamed givens to be in sync with p1673r13

iburyl:
givens_rotation functions were renamed at p1673r13


mhoemmen (approved):
Thanks Ilya! : - )

PR #294, opened 2024-10-04
fixed typo in conjugated implementation

iburyl:
mdspan has no .nested_accessor()


mhoemmen (approved):
Thanks for the fix! : - )

PR #303, opened 2025-01-24
Macos workflow

lucbv:
Adding an OSX CI workflow so that the work to get things working with AppleClang does not regress over time.


lucbv:
After a few trial and errors, this seems to do the job pretty well : )

masterleinad:
Can't uou just do something like https://stackoverflow.com/a/78173242? or do you actually need change more than just the underlying runner OS?

lucbv:
@masterleinad there are a few additional improvements made as you can see but we can revert to a more complicated and slower build option if you want?

crtrott (approved):

Issue #299, opened 2025-01-10
`extents` vs ` _extents`


prlw1:
When including `linalg` I see the following error:

```
In file included from /usr/src/local/stdBLAS/include/experimental/linalg:50,
                 from blas.cc:7:
/usr/src/local/stdBLAS/include/experimental/__p1673_bits/layout_blas_general.hpp: In member function 'constexpr Kokkos::Experimental::__p1673_version_0::linalg::__layout_blas_general_impl::__layout_blas_impl<BaseLayout, StaticLDA>& Kokkos::Experimental::__p1673_version_0::linalg::__layout_blas_general_impl::__layout_blas_impl<BaseLayout, StaticLDA>::operator=(const Kokkos::Experimental::__p1673_version_0::linalg::__layout_blas_general_impl::__layout_blas_impl<OtherExtents, OtherLDA>&)':
/usr/src/local/stdBLAS/include/experimental/__p1673_bits/layout_blas_general.hpp:133:11: error: 'class Kokkos::Experimental::__p1673_version_0::linalg::__layout_blas_general_impl::__layout_blas_impl<BaseLayout, StaticLDA>' has no member named '_extents'; did you mean 'extents'? [-Wtemplate-body]
  133 |     this->_extents = other.extents();
      |           ^~~~~~~~
      |           extents
```
Just in case, I commented out `#define LINALG_ENABLE_BLAS` in `__p1673_bits/linalg_config.h`, but no change...
(This is with `HEAD` `mdspan` and `stdBLAS`)

PR #305, opened 2025-05-20
layout_blas_general does not have an _extents member

lucbv:
Fix for issue #299 where we try to assign to _extents instead of assigning to _base_layout.extents


mhoemmen:
Btw, we replaced `layout_blas_general` in P1673 with `layout_left_padded` and `layout_right_padded` in P2642.

lucbv:
Ah that's good to know, this bug was preventing me from compiling the code so figured I would send a fix but I can modify my PR to just remove `layout_blas_general`.

mhoemmen:
@lucbv wrote:
> ...  I can modify my PR to just remove `layout_blas_general`

That would be delightful of you -- thank you! : - )

mhoemmen:
@youyu3 FYI

PR #306, opened 2025-05-23
Reviving Kokkos Kernels backend

lucbv:
Refreshing some of the implementation of the Kokkos Kernels backend so that we can re-compile and run the tests.
Mostly changing namespaces and a few signatures that changed, especially that of `std::extents`.
The mdspan_to_view test compiles and fails at runtime because the checks are not correct.
`utest_dot_kokkos_exe` test compiles and passes.

Note: something is not working with the complex test and I'm choosing to ignore it for the time being but if you have ideas for it let me know...

Let me know if you want the general approach I took to be changed or if modifying the other tests along the same lines would be reasonable?

### DIFF
--- a/tests/kokkos-based/helpers.hpp
+++ b/tests/kokkos-based/helpers.hpp
@@ -92,7 +92,7 @@ mdspan_t make_mdspan(ValueType *data, std::size_t ext0, std::size_t ext1) {
 template<class A_t, class ValueType = typename A_t::value_type>
 void set(A_t A, ValueType value)
 {
-  using size_type = typename Kokkos::extents<size_t, >::size_type;
+  using size_type = typename Kokkos::extents<size_t>::size_type;
   for (size_type i = 0; i < A.extent(0); ++i) {
     for (size_type j = 0; j < A.extent(1); ++j) {
       A(i, j) = value;
@@ -112,7 +112,7 @@ auto abs_max(mdspan<ElementType, extents<size_t, Extent>, LayoutPolicy, Accessor
   if (size == 0) {
     throw std::runtime_error("abs_max() requires non-empty input");
   }
-  const auto i = std::experimental::linalg::vector_idx_abs_max(v);
+  const auto i = MDSPAN_IMPL_STANDARD_NAMESPACE::MDSPAN_IMPL_PROPOSED_NAMESPACE::linalg::vector_idx_abs_max(v);
   if (i >= size) { // shouldn't happen: empty case is handled above
     throw std::runtime_error("Fatal: vector_idx_abs_max() failed");
   }
--- a/tests/kokkos-based/helpers.hpp
+++ b/tests/kokkos-based/helpers.hpp
@@ -18,8 +18,8 @@
 #ifndef LINALG_TESTS_KOKKOS_HELPERS_HPP_
 #define LINALG_TESTS_KOKKOS_HELPERS_HPP_
 
+#include <mdspan/mdspan.hpp>
 #include <experimental/linalg>
-#include <experimental/mdspan>
 #include <random>
 
 #if KOKKOS_VERSION < 30699
@@ -92,7 +92,7 @@ mdspan_t make_mdspan(ValueType *data, std::size_t ext0, std::size_t ext1) {
 template<class A_t, class ValueType = typename A_t::value_type>
 void set(A_t A, ValueType value)
 {
-  using size_type = typename std::experimental::extents<>::size_type;
+  using size_type = typename Kokkos::extents<size_t, >::size_type;
   for (size_type i = 0; i < A.extent(0); ++i) {
     for (size_type j = 0; j < A.extent(1); ++j) {
       A(i, j) = value;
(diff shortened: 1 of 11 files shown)


tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/mdspan_to_view_mapper_kk.hpp
```diff
@@ -124,14 +124,14 @@ auto mdspan_to_view(std::experimental::mdspan<
...
+  Kokkos::extents<size_t>::size_type ext1,
   class NestedLayout,
   class Accessor>
-auto mdspan_to_view(std::experimental::mdspan<
+auto mdspan_to_view(Kokkos::mdspan<
 		    ElementType,
-		    std::experimental::extents<ext0, ext1>,
-		    std::experimental::linalg::layout_transpose<NestedLayout>,
+		    Kokkos::extents<size_t, ext0, ext1>,
+		    MDSPAN_IMPL_STANDARD_NAMESPACE::MDSPAN_IMPL_PROPOSED_NAMESPACE::linalg::layout_transpose<NestedLayout>,
```
mhoemmen, line 134: This code uses the namespace macros, but the code above hard-codes "Kokkos" as the namespace.  Should we consider being more consistent?  Does enabling the kokkos-kernels back-end imply that `MDSPAN_IMPL_STANDARD_NAMESPACE` is always "Kokkos"?

tests/kokkos-based/helpers.hpp
```diff
@@ -92,7 +92,7 @@ mdspan_t make_mdspan(ValueType *data, std::size_t ext0, std::size_t ext1) {
 template<class A_t, class ValueType = typename A_t::value_type>
 void set(A_t A, ValueType value)
 {
-  using size_type = typename std::experimental::extents<>::size_type;
+  using size_type = typename Kokkos::extents<size_t>::size_type;
```
mhoemmen, line 95: `size_type` is `make_unsigned_t<index_type>`.  You probably want `index_type` throughout.

mhoemmen (changes requested):
Thanks for taking this up! : - )

I'm totally fine with merging this after some fixes that I discussed in the comments.  In particular, if you could please fix the non-kokkos-kernels test rather than commenting out those lines of code, that would be most excellent; thanks!

For the longer term, we should think about where projects like this should live.  The software dependency diagram has an uncomfortable cycle: KokkosCore depends on the C++ Standard Library, kokkos-kernels depends on KokkosCore, std::linalg depends on kokkos-kernels?!?

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/static_extent_match.hpp
```diff
@@ -24,8 +24,8 @@ namespace Impl {
 template <class size_type>
 constexpr bool static_extent_match(size_type extent1, size_type extent2)
 {
-  return extent1 == std::experimental::dynamic_extent ||
-         extent2 == std::experimental::dynamic_extent ||
+  return extent1 == Kokkos::dynamic_extent ||
+         extent2 == Kokkos::dynamic_extent ||
```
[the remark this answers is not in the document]
crtrott: Kokkos promises the mdspan types in the Kokkos namespace. In the future these may be aliases to std:: classes but they are not right now (we don't have a full C++26 compatible std implementation, even the C++23 ones which exist can't support GPUs). Since KokkosKernels depends on core I think it's fine to use the Kokkos namespace here.

tpl-implementations/include/experimental/__p1673_bits/kokkos-kernels/mdspan_to_view_mapper_kk.hpp
crtrott: FIrst this isn't actually a standard library implementation here. If (or maybe optimistically "when") we contribute this stuff I am not sure that we could move the KokkosKernels impl with it. But if we did the answer to the dependency thing is simple. std::linalg -> KokkosKernels -> KokkosCore -> std stuff not including linalg (i.e. mdspan and what not). These internal dependencies in the library are not uncommon. The same would be true for calling cuBLAS or oneMKL inside the standard implementation. Both of those also have a dependency on the standard library. Same is already true for stuff like using TBB inside the parallel algorithms implementation.

tests/kokkos-based/mdspan_to_view.cpp
```diff
@@ -46,7 +46,7 @@ void mdspan_to_view_test_impl()
   // rank2, non-const
   {
     std::vector<MDSpanValueType> a(12);
-    using mdspan_t = mdspan<MDSpanValueType, extents<dynamic_extent, dynamic_extent>>;
+    using mdspan_t = mdspan<MDSpanValueType, extents<size_t, dynamic_extent, dynamic_extent>>;
```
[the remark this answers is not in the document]
crtrott: Do we have it implemented?

PR #326, opened 2026-03-17
Improved continuous integration testing

amklinv-nnl:
Now supports Windows/MSVC as well as Linux/gcc and MacOS/clang. Also tests for both 17 and 20 standard compatibility.

### DIFF
--- a/.github/workflows/ghaction.yml
+++ b/.github/workflows/ghaction.yml
@@ -1,16 +1,13 @@
-name: stdBLAS
+name: build/test
 
 on: [push, pull_request]
-
-env:
-  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
-  BUILD_TYPE: RelWithDebInfo
 
 jobs:
   osmatrix: 
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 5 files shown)


amklinv-nnl:
The C++17 build _is_ broken. I'll submit a separate PR fixing it.

mhoemmen:
# Fix suggestions

## `proxy_reference.hpp`

```c++
friend auto abs(const derived_type& x) {
  if constexpr (std::is_unsigned_v<value_type>) {
    return value_type(static_cast<const this_type&>(x));
  } else {
    return abs(value_type(static_cast<const this_type&>(x))); // causes build error, possibly because it fails to exclude `std::abs` via the `= delete` idiom
  }
}
```

This code probably predates _`abs-if-needed`_.  We could try changing the implementation as follows.

```c++
friend auto abs(const derived_type& x) {
  return impl::abs_if_needed(value_type(static_cast<const this_type&>(x)));
}
```

amklinv-nnl:
@mhoemmen I implemented the changes you suggested, and the build/tests are passing. Kindly consider accepting this PR at your convenience!

mhoemmen (approved):
Thanks for the fixes! : - )

PR #330, opened 2026-05-14
Changed all instances of SizeType to IndexType

amklinv-nnl:
Closes #322


mhoemmen (approved):
Thanks for the fix! : - )

Issue #328, opened 2026-05-13
wrapping a vendor dscal


amklinv-nnl:
@georgemalerbo is interested in automatically generating BLAS wrappers for us. Thanks George!

Before we do that, @mhoemmen could you please provide feedback on the following dscal wrapper?
```c++
// Compiled with g++-16 -std=c++23 dscal_example.cpp -lblas
#include<cassert>
#include <vector>
#include <mdspan>
#include <iostream>

using namespace std;

// Should come from CMake or a config header
using blas_index_t = int; // BLAS typically uses int for sizes and indices

extern "C" {
    // We need to account for the Fortran name mangling, which often appends an underscore to the function name
    // The mangling should be automatically detected
    void dscal_(blas_index_t* n, double* a, double* x, blas_index_t* incx);
}

// Which accessors does it make sense to support?
template<class SizeType,
         ::std::size_t ext0,
         class Layout,
         class Accessor>
// scal requires strided data, so we can enforce this at compile time
// What is the difference between is_always_strided and is_strided? Should we support both?
requires ( mdspan<double, extents<SizeType, ext0>, Layout, Accessor>::is_always_strided() )
void scale(double a, mdspan<double, extents<SizeType, ext0>, Layout, Accessor> x) {
    blas_index_t n = x.extent(0);
    // This should be checked before calling the function
    // Like, in a "does it make sense to call BLAS?" function
    assert(n <= std::numeric_limits<blas_index_t>::max() && "Size of x exceeds maximum value for blas_index_t");

    // Assuming contiguous data, this will be 1 for a standard vector
    // This code will still work for non-contiguous data, but the performance may be worse due to cache misses
    blas_index_t incx = x.stride(0);
    dscal_(&n, &a, x.data_handle(), &incx);
}

int main() {
    vector<double> x_data = {1.0, 2.0, 3.0};
    auto x = mdspan(x_data.data(), x_data.size());

    double a = 2.0;
    scale(a, x);

    for(int i = 0; i < x.extent(0); ++i) {
        cout << x[i] << " ";
    }
    cout << endl;

    return 0;
}
```

mhoemmen:
Thanks @georgemalerbo for helping out! : - )

```
using namespace std;
```

Putting this declaration inside `main` would avoid exposing it to the rest of the program.

```
    // We need to account for the Fortran name mangling, which often appends an underscore to the function name
    // The mangling should be automatically detected
```

[...]

PR #331, opened 2026-05-20
Wrap scal

amklinv-nnl:
DO NOT MERGE.

For discussion in #328

### DIFF
--- a/include/experimental/__p1673_bits/blas1_scale.hpp
+++ b/include/experimental/__p1673_bits/blas1_scale.hpp
@@ -72,6 +72,7 @@ bool try_blas_scale(
   const Scalar alpha,
   mdspan<ElementType, extents<IndexType, ext0>, Layout, Accessor> x)
 {
+  #ifdef KOKKOS_ENABLE_CBLAS
   auto n = x.extent(0);
   auto incx = x.stride(0);
   if (x.is_strided() && n <= std::numeric_limits<CBLAS_INDEX>::max()) {
@@ -103,6 +104,7 @@ bool try_blas_scale(
       }
     }
   }
+  #endif
   return false;
 }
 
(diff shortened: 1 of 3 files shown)


include/experimental/__p1673_bits/blas1_scale.hpp
```diff
@@ -40,6 +43,90 @@ void linalg_scale_rank_1(
...
+         class Accessor,
+         class Scalar>
+bool try_blas_scale(
+  const Scalar alpha,
+  mdspan<ElementType, extents<IndexType, ext0>, Layout, Accessor> x)
+{
+  #ifdef KOKKOS_ENABLE_CBLAS
+  auto n = x.extent(0);
+  auto incx = x.stride(0);
+  if (x.is_strided() && n <= std::numeric_limits<CBLAS_INDEX>::max()) {
```
mhoemmen: 1. Per [[mdspan.layout.reqmnts] 19](https://eel.is/c++draft/views.multidim#mdspan.layout.reqmts-19), we can't call `x.stride(0)` until we know that `x.is_strided()` is `true`.

2. `incx` might be larger than `n`.

```suggestion
  if (x.is_strided() && n <= std::numeric_limits<int>::max()) {
    auto incx = x.stride(0);
    if (incx > std::numeric_limits<int>::max()) {
      return false;
    }
```

mhoemmen: It's also acceptable to turn this into a compile-time check.  We don't have to worry too much about weird layouts for which `is_strided()` is `true` sometimes but `is_always_strided()` is `false`.  

```
if constexpr (! x.is_always_strided()) {
  return false;
}
```

mhoemmen: Also, strided layout mappings can have stride zero.  `layout_stride` cannot, unless its corresponding extent is also zero.  If `incx` is zero, then we should also return `false`.

mhoemmen (changes requested):

PR #333, opened 2026-06-29
Added codecov integration to stdBLAS

amklinv-nnl:
Closes NavalNuclearLab/stdBLAS#7

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,93 @@
+name: Coverage
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  coverage:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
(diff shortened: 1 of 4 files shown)


amklinv-nnl:
@mhoemmen You can view the coverage report here: https://app.codecov.io/github/kokkos/stdBLAS/pull/333/tree/include/experimental/__p1673_bits?dropdown=coverage

The numbers are not particularly meaningful to us, especially given how code coverage deals with templates and ifdefs. The goal is that for future commits, we can get an idea of whether the code being introduced is actually being executed by the tests. Example: if we introduce parallel scale, did that version of the code ever get executed, or did we fail to test it because LLVM doesn't support std::execution::par yet? Did we test our vendor BLAS wrappers, or did we fail to because the github runner didn't have a BLAS implementation installed?

mhoemmen (approved):

PR #334, opened 2026-07-30
Fix stride(0) ordering, bound incx, reject incx == 0 and deduce the CBLAS index type

georgemalerbo:
Applies what @mhoemmen suggested in kokkos/stdBLAS#331 x.stride(0) is now only called after x.is_strided() is confirmed, incx is bounded by the index type's max, and incx == 0 returns false so generic fallback handles it.

Also implemented the deduction of the BLAS integer index type discussed in kokkos/stdBLAS#328: a variadic function template declaration, used only inside decltype, extracts the type of cblas_dscal's first parameter

### DIFF
--- a/include/experimental/__p1673_bits/blas1_scale.hpp
+++ b/include/experimental/__p1673_bits/blas1_scale.hpp
@@ -72,6 +72,7 @@ bool try_blas_scale(
   const Scalar alpha,
   mdspan<ElementType, extents<IndexType, ext0>, Layout, Accessor> x)
 {
+  #ifdef KOKKOS_ENABLE_CBLAS
   auto n = x.extent(0);
   auto incx = x.stride(0);
   if (x.is_strided() && n <= std::numeric_limits<CBLAS_INDEX>::max()) {
@@ -103,6 +104,7 @@ bool try_blas_scale(
       }
     }
   }
+  #endif
   return false;
 }
 
--- a/include/experimental/__p1673_bits/blas1_scale.hpp
+++ b/include/experimental/__p1673_bits/blas1_scale.hpp
@@ -18,6 +18,9 @@
 #ifndef LINALG_INCLUDE_EXPERIMENTAL___P1673_BITS_BLAS1_SCALE_HPP_
 #define LINALG_INCLUDE_EXPERIMENTAL___P1673_BITS_BLAS1_SCALE_HPP_
 
+#include "blas_helpers.hpp"
+#include <type_traits>
+
 namespace MDSPAN_IMPL_STANDARD_NAMESPACE {
 namespace MDSPAN_IMPL_PROPOSED_NAMESPACE {
 inline namespace __p1673_version_0 {
@@ -31,12 +34,94 @@ template<class ElementType,
          class Layout,
          class Accessor,
          class Scalar>
-void linalg_scale_rank_1(
+void generic_scale_rank_1(
   const Scalar alpha,
   mdspan<ElementType, extents<IndexType, ext0>, Layout, Accessor> x)
 {
   for (IndexType i = 0; i < x.extent(0); ++i) {
     x(i) *= alpha;
+  }
+}
+
+template<class Scalar, class MdspanType>
+constexpr bool maybe_can_blas_scale() {
+  // It's OK if Scalar and value_type aren't the same,
+  // as long as we can convert Scalar to value_type.
+  using value_type = typename MdspanType::value_type;
+  constexpr bool blas_value_type =
+    std::is_convertible_v<Scalar, value_type> &&
+    impl::is_blas_value_type_v<value_type>;
+
+  constexpr bool blas_layout =
+    impl::is_blas_layout_type_v<typename MdspanType::layout_type>;
+
+  constexpr bool blas_accessor =
+    impl::is_blas_accessor_type_v<typename MdspanType::accessor_type>;
+
+  return blas_value_type && blas_layout && blas_accessor;
+} 
+
+// Return true if the BLAS call was successfull, false otherwise.
+template<class ElementType,
+	 class IndexType,
+         ::std::size_t ext0,
+         class Layout,
+         class Accessor,
+         class Scalar>
+bool try_blas_scale(
+  const Scalar alpha,
+  mdspan<ElementType, extents<IndexType, ext0>, Layout, Accessor> x)
+{
+  auto n = x.extent(0);
+  auto incx = x.stride(0);
+  if (x.is_strided() && n <= std::numeric_limits<CBLAS_INDEX>::max()) {
+    if constexpr (std::is_same_v<ElementType, float>) {
+      cblas_sscal(n, alpha, x.data_handle(), incx);
+      return true;
+    } else if constexpr (std::is_same_v<ElementType, double>) {
+      cblas_dscal(n, alpha, x.data_handle(), incx);
+      return true;
+    } else if constexpr (std::is_same_v<ElementType, std::complex<float>>) {
+      if constexpr(std::is_convertible_v<Scalar, float>) {
+        cblas_csscal(n, alpha, x.data_handle(), incx);
+        return true;
+      }
+      else if constexpr (std::is_convertible_v<Scalar, std::complex<float>>) {
+        auto converted_alpha = static_cast<std::complex<float>>(alpha);
+        cblas_cscal(n, &converted_alpha, x.data_handle(), incx);
+        return true;
+      }
+    } else if constexpr (std::is_same_v<ElementType, std::complex<double>>) {
+      if constexpr (std::is_convertible_v<Scalar, double>) {
+        cblas_zdscal(n, alpha, x.data_handle(), incx);
+        return true;
+      }
+      else if constexpr (std::is_convertible_v<Scalar, std::complex<double>>) {
+        auto converted_alpha = static_cast<std::complex<double>>(alpha);
+        cblas_zscal(n, &converted_alpha, x.data_handle(), incx);
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+template<class ElementType,
+	 class IndexType,
+         ::std::size_t ext0,
+         class Layout,
+         class Accessor,
+         class Scalar>
+void linalg_scale_rank_1(
+  const Scalar alpha,
+  mdspan<ElementType, extents<IndexType, ext0>, Layout, Accessor> x)
+{
+  bool done = false;
+  if constexpr (maybe_can_blas_scale<Scalar, decltype(x)>()) {
+    done = try_blas_scale(alpha, x);
+  }
+  if (!done) {
+    generic_scale_rank_1(alpha, x);
   }
 }
 
--- a/include/experimental/__p1673_bits/blas1_scale.hpp
+++ b/include/experimental/__p1673_bits/blas1_scale.hpp
@@ -75,7 +75,7 @@ bool try_blas_scale(
   #ifdef KOKKOS_ENABLE_CBLAS
   auto n = x.extent(0);
   auto incx = x.stride(0);
-  if (x.is_strided() && n <= std::numeric_limits<CBLAS_INDEX>::max()) {
+  if (x.is_strided() && n <= std::numeric_limits<int>::max()) {
     if constexpr (std::is_same_v<ElementType, float>) {
       cblas_sscal(n, alpha, x.data_handle(), incx);
       return true;
--- a/include/experimental/__p1673_bits/blas1_scale.hpp
+++ b/include/experimental/__p1673_bits/blas1_scale.hpp
@@ -74,8 +74,16 @@ bool try_blas_scale(
 {
   #ifdef KOKKOS_ENABLE_CBLAS
   auto n = x.extent(0);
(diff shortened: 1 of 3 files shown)


include/experimental/__p1673_bits/blas_helpers.hpp
```diff
@@ -0,0 +1,103 @@
...
+  // This just means that the layouts COULD be valid.
+  // For layout_stride, we need to check the strides first.
+  std::is_same_v<Layout, layout_left> ||
+  std::is_same_v<Layout, layout_right> ||
+  // AMK 5/20/26 - layout_left_padded and layout_right_padded were added in C++26.
+  // According to cppreference, padded layouts are covered by the same feature test as submdspan.
+  #ifdef __cpp_lib_submdspan
+  std::is_same_v<Layout, layout_left_padded> ||
+  std::is_same_v<Layout, layout_right_padded> ||
+  #endif
```
mhoemmen: `#include <mdspan/mdspan.hpp>` will pull in the reference implementation of mdspan.  That doesn't use the Standard feature test macros.  Thus, this part of the code would never get tested with the current CI workflow.

In fact, this code is not well formed, because `layout_left_padded` and `layout_right_padded` are class templates that take a `size_t` constant template parameter, unlike `layout_left` and `layout_right` that are ordinary class types without a template parameter.

Here's one way to fix this.

1. Given that we've only ever tested with the reference mdspan implementation, we could simply remove the macro test.
2. Define `is_padded_layout_v` traits for any specializations of `layout_left_padded` and `layout_right_padded`.
3. Use `is_padded_layout_v` here.

It would take some header file and macro refactoring for the reference std::linalg implementation to be able to use the actual `<mdspan>` header from an implementation like GCC's or Clang's.

include/experimental/__p1673_bits/blas1_scale.hpp
```diff
@@ -40,6 +43,98 @@ void linalg_scale_rank_1(
...
+  constexpr bool blas_layout =
+    impl::is_blas_layout_type_v<typename MdspanType::layout_type>;
+
+  constexpr bool blas_accessor =
+    impl::is_blas_accessor_type_v<typename MdspanType::accessor_type>;
+
+  return blas_value_type && blas_layout && blas_accessor;
+} 
+
+// Return true if the BLAS call was successfull, false otherwise.
```
mhoemmen:
```suggestion
// Return true if a BLAS routine could be called to scale the vector, false otherwise.
```

The comment suggests that the BLAS could be called unsuccessfully, but the BLAS functions that could be called here don't have a failure mode.

mhoemmen (changes requested):
Thanks for the contribution! : - )

The code inside `#ifdef __cpp_lib_submdspan` ... `#endif` is not well formed.  This suggests that we don't have test coverage for that case.  Would you consider adding tests for `layout_left_padded` and `layout_right_padded`?  The tests should not be gated on `__cpp_lib_submdspan`, for the reasons explained in the comment on that part of the code.  Thanks!

georgemalerbo:
I've been working through the changes you requested, but ended up pushing more to my fork's branch than I meant to, some of it was experimentation that really belongs on a separate branch. I've force pushed the branch back to 6a94735, the commit you reviewed.

mhoemmen:
@georgemalerbo wrote:
> I've been working through the changes you requested, but ended up pushing more to my fork's branch than I meant to, some of it was experimentation that really belongs on a separate branch. I've force pushed the branch back to https://github.com/kokkos/stdBLAS/commit/6a9473572ec6f856897779a3b83057ffa99009e8, the commit you reviewed.

Thanks for working on this!  Did you have any questions about the requested changes?

georgemalerbo:
- [110831f3](https://github.com/kokkos/stdBLAS/pull/334/commits/110831f3c68871c23a3b75a9eb74dd9b31bc50d5) drops the `__cpp_lib_submdspan` test and adds `is_padded_layout_v`
- [88d7d1fc](https://github.com/kokkos/stdBLAS/pull/334/commits/88d7d1fcac6aab521be8c9e5b8c310b2c99fb5a1) adds the padded layout and `layout_stride` tests, ungated
- [1e6c8816](https://github.com/kokkos/stdBLAS/pull/334/commits/1e6c881653b3f6a60e5d09031ecf8197ab936c40) applies your wording for the `try_blas_scale` comment

- [fc92d10d](https://github.com/kokkos/stdBLAS/pull/334/commits/fc92d10dc75148040a58c0722d1b2050cd7d9b8c) - Allow parent project to set LINALG_ options and cache variables

Rather than raise the minimum cmake we can keep it at 3.12 and opt into CMP0077 and CMP0126, so a user of stdBLAS with a version of cmake 3.21 or newer can configure their project easily:

```cmake
set(LINALG_ENABLE_BLAS ON)
set(LINALG_CXX_STANDARD 20)
add_subdirectory(stdBLAS)
```

- [2c7bb8d9](https://github.com/kokkos/stdBLAS/pull/334/commits/2c7bb8d91bfcade8bd6bdf6a2a1aed7ddf053244) - Export the generated linalg_config.h from the linalg target

`include_directories()` covered the project's own tests and examples, but not a user/parent target and the `linalg` target exported only the source include directory

- [feb8802f](https://github.com/kokkos/stdBLAS/pull/334/commits/feb8802fd391732830c98f94e66540b4d1052df8) - Link mdspan by its exported target name and require it from the installed package

`target_link_libraries(linalg INTERFACE mdspan)` worked as long as mdspan came from FetchContent with its plain `mdspan` target and but breaks if someone tries to use an installed mdspan via `find_package(mdspan)` not from the in tree source build that FetchContent populates where the package config exports `mdspan::mdspan` instead

- [6827f449](https://github.com/kokkos/stdBLAS/pull/334/commits/6827f4494cea6cb1629ff0ca98be7d20cd575741) - Include mdspan from experimental/linalg

`<experimental/linalg>` opens `MDSPAN_IMPL_STANDARD_NAMESPACE` without including `<mdspan/mdspan.hpp>` so it only compiles when something else included mdspan first which the tests do but a user who includes it on its own does not

A user doing the ordinary thing against the tree as it was before this commit:

```cmake
cmake_minimum_required(VERSION 3.12)
project(app CXX)

add_subdirectory(stdBLAS)

add_executable(app main.cpp)
target_link_libraries(app linalg)
```

```cpp
// main.cpp
#include <experimental/linalg>

int main() {}
```

```console
$ cmake -S . -B build && cmake --build build
In file included from stdBLAS/include/experimental/linalg:35,
                 from main.cpp:1:
stdBLAS/include/experimental/__p1673_bits/blas1_givens.hpp:81:16: error: ‘is_inline_exec_v’ is not a member of ‘Kokkos::Experimental::__p1673_version_0::linalg::impl’
   81 |     && ! impl::is_inline_exec_v<Exec>
      |                ^~~~~~~~~~~~~~~~
```

and with this commit in, the same project unchanged:

```console
$ cmake -S . -B build && cmake --build build
-- The CXX compiler identification is GNU 13.3.0
-- Detected support for C++23 standard
-- Using "-fconcepts" to enable concepts support
-- No installed mdspan found, fetching from Github
-- Build include directory: build/stdBLAS/include/experimental
-- Configuring done (2.9s)
-- Generating done (0.0s)
-- Build files have been written to: build
[ 50%] Building CXX object CMakeFiles/app.dir/main.cpp.o
[100%] Linking CXX executable app
[100%] Built target app
```

- [fc12c109](https://github.com/kokkos/stdBLAS/pull/334/commits/fc12c10979105fc3c07cefa6e0c1404c2ccd7038) - Define LINALG_HAS_CBLAS in linalg_config.h

- adds `#cmakedefine LINALG_HAS_CBLAS` to `linalg_config.h.in`
- includes that header where the guard is tested
- tests `LINALG_HAS_CBLAS` in place of `KOKKOS_ENABLE_CBLAS`

`LINALG_ENABLE_BLAS` looks settled already so I kept it. Renaming it to `LINALG_ENABLE_CBLAS` was my first thought since calling CBLAS is what it actually enables. I went with a `HAS` name for the internal macro rather than another `ENABLE` because `LINALG_ENABLE_BLAS` is what the user asks for and `LINALG_HAS_CBLAS` is what the build got

- [f92109ce](https://github.com/kokkos/stdBLAS/pull/334/commits/f92109ceaeec9762d9d3bdcdd905eeda52348061) Find and link a BLAS behind LINALG_ENABLE_BLAS

- keeps whatever `find_package(BLAS)` reports instead of overwriting it
- makes `LINALG_ENABLE_BLAS` explicit
- carries the BLAS on the `linalg` target as an `INTERFACE` requirement, where before it reached the tests only and not the `linalg` itself so no one using library got a BLAS to link against

A user who adds `add_subdirectory(stdBLAS)` and asks for nothing now gets `LINALG_ENABLE_BLAS:BOOL=OFF` instead of `ON`. A BLAS is still found by `find_package(BLAS)` the presence of a `cblas.h` just no longer decides anything.

- [4b500ff0](https://github.com/kokkos/stdBLAS/pull/334/commits/4b500ff0ceb984ca95a22520998b7d30d5b25e87) Let a user name the CBLAS header when the provider does not call it cblas.h

- adds `LINALG_CBLAS_HEADER` and `LINALG_CBLAS_INCLUDE_DIRS` cache variables
- carries the chosen name into `linalg_config.h` as the `#include` argument, where before it was hardcoded to `cblas.h`, so every file that includes stdBLAS uses the header that was named

- [1786baf0](https://github.com/kokkos/stdBLAS/pull/334/commits/1786baf0266aedd054dc643e504fd8955c8d7f76) Document LINALG_ENABLE_BLAS

States that what it enables is calling CBLAS, and that the header can be named when the provider does not call it `cblas.h`.

- [4189f5eb](https://github.com/kokkos/stdBLAS/pull/334/commits/4189f5eb9b0ae79833d6c118486324dfb84b2f90) Build and test the CBLAS path in CI

- installs `libopenblas-dev`, turns `LINALG_ENABLE_BLAS` on and runs the tests at C++17 and C++20
- asserts the generated `linalg_config.h` contains `#define LINALG_HAS_CBLAS`
- asserts the test binary has an undefined `cblas_dscal` to resolve

include/experimental/__p1673_bits/blas_helpers.hpp
georgemalerbo: Done in [110831f3](https://github.com/kokkos/stdBLAS/pull/334/commits/110831f3c68871c23a3b75a9eb74dd9b31bc50d5): drop the `__cpp_lib_submdspan` test, add an `is_padded_layout_v` trait that matches any specialization of `layout_left_padded` and `layout_right_padded`, and use that instead.

I think `__cpp_lib_aligned_accessor` has the same problem
If  `aligned_accessor` is used without  `std::`, the name is looked up in Kokkos and not std so reference mdspan has no `aligned_accessor` so `blas_helpers.hpp` breaks and `<experimental/linalg>` is unincludable

So also removed  `__cpp_lib_aligned_accessor` ,  `is_aligned_accessor_v` and its term in `is_blas_accessor_type_v`.


Build 6a94735 with `-D__cpp_lib_aligned_accessor=202411L` :

```
In file included from include/experimental/__p1673_bits/blas1_scale.hpp:21,
                 from include/experimental/linalg:40,
                 from tests/native/./gtest_fixtures.hpp:31,
                 from tests/native/scale.cpp:1:
include/experimental/__p1673_bits/blas_helpers.hpp:77:38: error: ‘aligned_accessor’ was not declared in this scope; did you mean ‘is_aligned_accessor_v’?
   77 | constexpr bool is_aligned_accessor_v<aligned_accessor<ElementType, ByteAlignment>> = true;
      |                                      ^~~~~~~~~~~~~~~~
      |                                      is_aligned_accessor_v
include/experimental/__p1673_bits/blas_helpers.hpp:77:16: error: parse error in template argument list
   77 | constexpr bool is_aligned_accessor_v<aligned_accessor<ElementType, ByteAlignment>> = true;
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
include/experimental/__p1673_bits/blas_helpers.hpp:77:16: error: wrong number of template arguments (2, should be 1)
include/experimental/__p1673_bits/blas_helpers.hpp:73:16: note: provided for ‘template<class Accessor> constexpr const bool Kokkos::Experimental::__p1673_version_0::linalg::impl::is_aligned_accessor_v<Accessor>’
   73 | constexpr bool is_aligned_accessor_v = false;
      |                ^~~~~~~~~~~~~~~~~~~~~
```

include/experimental/__p1673_bits/blas1_scale.hpp
georgemalerbo: Done in [1e6c8816](https://github.com/kokkos/stdBLAS/pull/334/commits/1e6c881653b3f6a60e5d09031ecf8197ab936c40)

mhoemmen:
@georgemalerbo The original title of this PR was "Fix stride(0) ordering, bound incx, reject incx == 0 and deduce the CBLAS index type."  You've done that, no?  Why not call that victory, get the PR merged, and create a new PR with the CI changes?  What motivates combining those previous changes with the CI changes into a single PR?

Reviewers need to be able to understand what's going on.  It's easier to do that if each PR fixes one thing.

